PR #24, opened 2023-06-06
Sc 5407/publish trello package

johnathan-rodriguez:
Made changes to the cli arguments to be more consistent and also how the package is being imported


johnathan-rodriguez:
Made a fresh branch to handle conflicts

PR #29, opened 2023-06-09
Sc 5406/asana library code

wrp801:
No library necessary as there is already an existing Asana package in PyPI. Two CLI's implemented for the application:
- create_task
- edit_task


shipyard_blueprints/asana/shipyard_asana/cli/edit_task.py
```diff
@@ -0,0 +1,68 @@
...
+
+EXIT_CODE_BAD_REQUEST = 201
+EXIT_CODE_TICKET_NOT_FOUND = 202
+
+def get_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--access-token',type=str, required = True, dest = 'access_token')
+    parser.add_argument('--ticket-id', type = str, required = True, dest = 'ticket_id')
+    parser.add_argument('--name', type = str, dest = 'name', required = False, default = '')
+    parser.add_argument('--resource-subtype', type = str, dest = 'resource_subtype', required = False, default = 'default_task', choices = {'default_task', 'milestone','section','approval'})
```
johnathan-rodriguez, line 15: for optional arguments with choices I believe we have to include an empty string as a choice. If it is left blank when the vessel runs i believe it will throw an error.

johnathan-rodriguez (commented):
Can you confirm that the optional arguments with choices are able to run in shipyard if left blank. You may have to add an empty string as an option

shipyard_blueprints/asana/shipyard_asana/cli/edit_task.py
johnathan-rodriguez: wouldnt argparse fail for an invalid choice of `''`  when no choice is made? [](edited-with-Axolo)

johnathan-rodriguez (approved):

shipyard_blueprints/asana/shipyard_asana/cli/edit_task.py
johnathan-rodriguez: approved [](edited-with-Axolo)

PR #31, opened 2023-06-13
Microsoft Teams Send Message [sc-5216]

johnathan-rodriguez:


shipyard_blueprints/microsoft-teams/shipyard_microsoft_teams/cli/post_message.py
wrp801, line 1: Code is fine, only thing I would add would be some helpful logging statements like we do in the others. For instance, if it fails we should try to log the message returned so that the user has some useful information as to what went wrong. Look at how I did the Asana CLI's and just copy the function to make the logger. Nothing else needs to be changed, just add that logger and swap out the print statements

wrp801 (changes requested):

PR #33, opened 2023-06-13
Added authentication test

wrp801:
Added authorization test


johnathan-rodriguez (approved):
I wouldve simplified this by at least removing the get_args. the variables could help with readability but personally i wouldve went with something like the below
```
def main():
     resp = requests.get(
          url = ...,
          headers = {
          "Authorization: f"Bearer {os.getenv('ASANA_ACCESS_TOKEN')}"
          }
     )
...
```
approved with food for thought if you would like to simply it

PR #35, opened 2023-06-21
Fivetran Library Code [sc-5557]

johnathan-rodriguez:


shipyard_blueprints/fivetran/shipyard_fivetran/fivetran.py
```diff
@@ -1,33 +1,272 @@
...
+        Raises:
+            ExitCodeException: If an error occurs while triggering sync.
+        """
+        if wait_for_completion:
+            self.logger.info(
+                "Getting connection details for last successful and failed syncs..."
+            )
+            prev_success, prev_failure = self._get_latest_success_and_failure(
+                connector_id
+            )
```
wrp801, line 114: Any reason why you wouldn't put the block from below right here since you are handling the `wait_for_completion` check? 

```python
                new_success, new_failure = prev_success, prev_failure
                while prev_success == new_success and prev_failure == new_failure:
                    self.logger.info("Waiting for sync to complete...")
                    time.sleep(poke_interval)
                    new_success, new_failure = self._get_latest_success_and_failure(
                        connector_id
                    )
                self.logger.info("Sync completed")
                self.logger.info("Checking for new failure")
                if (
                    prev_failure
                    and new_failure != prev_failure
                    or not prev_failure
                    and new_failure
                ):
                    raise ExitCodeException(
                        f"Sync failed at {new_failure}",
                        self.EXIT_CODE_SYNC_REFRESH_ERROR,
                    )
                else:
                    self.logger.info("No new failure detected")

```

wrp801 (changes requested):

shipyard_blueprints/fivetran/shipyard_fivetran/fivetran.py
johnathan-rodriguez: I wanted a snapshot of the values before the sync was attempted in the unlikely scenario there is an edge case that can mark the succeeded_at or failed at almost instantly

wrp801 (approved):

PR #44, opened 2023-06-29
Moved the Census blueprint to the mono repo and adjusted the cli to a…

johnathan-rodriguez:
…llow for an optional wait parameter


shipyard_blueprints/census/pyproject.toml
```diff
@@ -2,14 +2,14 @@
...
+#readme = "README.md"
 packages = [{include = "shipyard_census"}]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-requests = "2.28.0"
+requests = "2.31.0"
 
 [tool.poetry.group.dev.dependencies]
 shipyard-bp-utils = {path = "../shipyard-bp-utils"}
```
wrp801, line 15: Add `develop = true` to any path dependency

wrp801 (changes requested):
Overall looks good, just a tiny change to make in the toml file

wrp801:
@johnathan-rodriguez any reason why you deleted the check_sync_cli.py file?

johnathan-rodriguez:
@wrp801 I thought we were removing the check sync and combining it with the trigger sync cli- do we want both for the new way?

johnathan-rodriguez:
the old check sync would exist in the old way of doing things so the legacy function should be intact

wrp801:
Okay so what you’re proposing is to leave the check syncs tied to the old repos?

johnathan-rodriguez:
yes

wrp801 (approved):

PR #49, opened 2023-07-25
Unit tests for the authtest.py

johnathan-rodriguez:


wrp801 (changes requested):
The Domo package needs to be to also authenticate with just an access token (and the domo instance I believe). Essentially we have to program it to expect either a:
- Client ID and Secret
OR 
- Access token and  Domo instance

wrp801 (approved):

PR #59, opened 2023-07-31
Adjust the cli command to allow for the wait for completion option

johnathan-rodriguez:


shipyard_blueprints/hightouch/pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [{include = "shipyard_hightouch"}]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-requests = "2.28.0"
```
wrp801, line 12: Add `shipyard-templates` to both the dev dependency group (do it as a path in that case i.e. `../../shipyard-templates`) and to the main dependency group. You may also need to add shipyard-utils to the main dependency group

wrp801 (changes requested):

wrp801:
You still need to add shipyard-templates to the main dependency group in the pyproject.toml file for Hightouch

wrp801 (approved):

PR #61, opened 2023-08-04
Coalesce add error status

johnathan-rodriguez:


johnathan-rodriguez:
Made a new PR with only the relevant commits

PR #63, opened 2023-08-08
Dockerfile for local development [sc-5859]

johnathan-rodriguez:
Included some of the premature hubspot code to be able to test the versioning and deployment github action next.


wrp801 (changes requested):
The `authtest_test.py` should be in the `tests` directory, not the `cli` directory

johnathan-rodriguez:
> The `authtest_test.py` should be in the `tests` directory, not the `cli` directory

To clarify, authtest_test.py is housed within the tests directory. However, I introduced a cli subdirectory inside the tests directory.The idea was to keep the main CLI function tests separate from the class method tests. If this setup seems off, let me know and I can put all tests under the same directory

PR #64, opened 2023-08-08
Sc 5858/GitHub action linter

johnathan-rodriguez:


wrp801:
This seems to be just a ton of formatting changes, plus the docker file, plus the hubsot code, plus the GH action correct?

johnathan-rodriguez:
Correct

johnathan-rodriguez:
I won’t push the packages so the changes should be esthetic it’s to have the build pass going forward

wrp801 (approved):

wrp801:
I approved this but I realized we don’t need a python 3.7 linter, just 3.9

johnathan-rodriguez:
Everything in the mono repo should be 3.9 but since we have legacy code in 3.7 I did it as a precaution. Also serves as a proof of concept for when/if we move to 3.11

PR #69, opened 2023-08-29
Hubspot Library Code [sc-5978]

johnathan-rodriguez:


johnathan-rodriguez:
tests failing due to the shipyard templates not being publish

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot.py
```diff
@@ -85,3 +102,184 @@ def connect(self):
         else:
             self.logger.info("Successfully connected to Hubspot API")
             return 0
+
+    def export_data(self, export_type: str, **kwargs):
+        export_type = validate_export_type(export_type)
+        self.logger.debug("Exporting data from Hubspot API")
+        if export_type == "list":
+            return self.export_list(**kwargs)
+        # elif export_type == "view":
```
johnathan-rodriguez, line 111: Opted out of including view export since there is no easy way to have this in a user friendly format in our UI. 
it takes search conditions as an array where as the list can be premade in hubspot and referred to in the blueprint.

shipyard_blueprints/hubspot/pyproject.toml
```diff
@@ -21,6 +21,7 @@ black = {version = "^23.7.0", optional = true}
 pytest = "^7.4.0"
 python-dotenv = "^1.0.0"
 pytest-dotenv = "^0.5.2"
+requests-mock = "^1.11.0"
 
 [build-system]
```
wrp801, line 26: You will need to add `shipyard-templates` to the main dependency group as well.

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot.py
```diff
@@ -1,34 +1,74 @@
...
+            for handler in self.logger.handlers:
+                handler.setLevel(DEBUG)
 
-    def _requests(self, endpoint: str, method: str = "GET", payload: dict = None):
+    def _requests(
+        self,
+        endpoint: str,
+        method: str = "GET",
+        payload: Optional[Dict] = None,
+        headers: Optional[Dict] = None,
```
wrp801, line 49: Any reason why we don't set the headers as an attribute for the class?

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot.py
```diff
@@ -85,3 +102,184 @@ def connect(self):
         else:
             self.logger.info("Successfully connected to Hubspot API")
             return 0
+
+    def export_data(self, export_type: str, **kwargs):
+        export_type = validate_export_type(export_type)
+        self.logger.debug("Exporting data from Hubspot API")
```
wrp801, line 108: Can you provide some doc strings as to what this method does and what are valid options for the `export_type` as well as valid key word args?

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot.py
```diff
@@ -85,3 +102,184 @@ def connect(self):
...
+            import_job_id = response["id"]
+            self.logger.info(
+                f"Successfully triggered import with import id: {import_job_id}"
+            )
+            return response
+
+        else:
+            handle_request_errors(response)
+
+    def import_contact_data(
```
wrp801, line 241: Same - provide docstrings here

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot_utils.py
```diff
@@ -0,0 +1,196 @@
+import os
+import json
+
+from shipyard_templates import Crm, ExitCodeException
+
+
+def handle_request_errors(response):
```
wrp801, line 7: Same - provide docstrings here

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot_utils.py
```diff
@@ -0,0 +1,196 @@
...
+            headers = file.readline().strip().split(",")
+
+            for header in headers:
+                if header == hubspot_alternate_id:
+                    mapping.append(
+                        column_to_hubspot(
+                            header, header, column_type="HUBSPOT_ALTERNATE_ID"
+                        )
+                    )
+                else:
```
wrp801, line 188: Please provide basic docstrings for all the functions in hubspot_utils.py

wrp801 (changes requested):

shipyard_blueprints/hubspot/shipyard_hubspot/hubspot.py
johnathan-rodriguez: I believe keeping it in the helper function makes it slightly more flexible. Not all of the endpoints accept the same headerdata- For example the import endpoint would fail if  tried to pass in the default header. it is a bad example considering that i opted to bypass the helper function but is an example of the headers needing some flexibility. In hindsight i should set the default value in the function instead of handling it within the script.

Not sure what the benefit would be to change it into the class attribute.

wrp801 (approved):

PR #72, opened 2023-09-15
Salesforce Library Code [sc-5987]

johnathan-rodriguez:


shipyard_blueprints/salesforce/shipyard_salesforce/salesforce.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        Verify that the Salesforce credentials are valid.
+
+        Returns:
+        - 0 if the credentials are valid
+        - 1 if the credentials are invalid
+        """
+        self.logger.info("Attempting to verify Salesforce credentials")
+        try:
+            self._request("")
```
wrp801, line 106: As currently constituted, does this return success?

shipyard-templates/shipyard_templates/exit_code_exception.py
```diff
@@ -1,3 +1,17 @@
+def standardize_to_exit_code_exception(func):
```
wrp801, line 1: I would rename this to something more succinct like `standardize_errors`

shipyard_blueprints/salesforce/shipyard_salesforce/salesforce_utils.py
```diff
@@ -0,0 +1,112 @@
+import json
+
+from typing import Optional
+from requests import Response
+from shipyard_templates import Crm, ExitCodeException
+
+
+def handle_request_errors(response: Response) -> None:
```
wrp801, line 8: Do you think something like this function could be used across other CRM vendors? Or is this particularly specific to Salesforce?

wrp801 (changes requested):

shipyard_blueprints/salesforce/shipyard_salesforce/salesforce.py
johnathan-rodriguez: yes- returns salesforce account metadata

shipyard_blueprints/salesforce/shipyard_salesforce/salesforce_utils.py
johnathan-rodriguez: this would be useful for vendors that have a significant number of error statuses

wrp801 (approved):

PR #75, opened 2023-09-24
Snowpark functionality

wrp801:


johnathan-rodriguez:
Can you run black and review the pytest so that the pipelines don't fail going forward

shipyard_blueprints/snowflake/README.md
```diff
@@ -0,0 +1,96 @@
...
+- [Installation](#installation)
+- [Usage](#usage)
+  - [SnowflakeClient](#snowflakeclient)
+  - [SnowparkClient](#snowparkclient)
+## Installation
+
+You can install `shipyard-snowflake` using `pip`:
+
+```bash
+pip install shipyard-snowflake
```
johnathan-rodriguez, line 16: I believe this should be `pip3`

shipyard_blueprints/snowflake/shipyard_snowflake/snowflake.py
```diff
@@ -148,3 +148,39 @@ def fetch(
...
+        self,
+        conn: snowflake.connector.SnowflakeConnection,
+        file_path: str,
+        table_name: str,
+    ):
+        """Executes a PUT command to load a file to internal staging. This is the fastest way to load a large file and should be followed by a copy into command
+
+        Args:
+            conn: The established Snowflake Connection
+            file: The file to load
```
johnathan-rodriguez, line 162: Update file to file_path to match naming in method

shipyard_blueprints/snowflake/shipyard_snowflake/snowflake.py
```diff
@@ -148,3 +148,39 @@ def fetch(
...
+            table_name: The table in Snowflake to write to
+        """
+        put_statement = f"""PUT file://{file_path}* '@%"{table_name}"' """
+        self.execute_query(conn, put_statement)
+
+    def copy_into(self, conn: snowflake.connector.SnowflakeConnection, table_name: str):
+        """Executes a COPY INTO command to load a file from internal staging to a table"""
+        copy_statement = f"""COPY INTO "{table_name}" FROM '@%\"{table_name}\"' PURGE=TRUE FILE_FORMAT=(TYPE=CSV FIELD_DELIMITER=',' COMPRESSION=GZIP, PARSE_HEADER=TRUE) MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE"""
```
johnathan-rodriguez, line 170: break this string up to be in multi lines(may be caught with black)

shipyard_blueprints/snowflake/shipyard_snowflake/snowflake.py
```diff
@@ -148,3 +148,39 @@ def fetch(
...
+        self.execute_query(conn, copy_statement)
+
+    def _create_table(
+        self,
+        table_name: str,
+        columns=None,
+    ) -> str:
+        """Returns the SQL for to create or replace a table in Snowflake
+        Args:
+            conn (snowflake.connector.SnowflakeConnection): The snowflake connection object
```
johnathan-rodriguez, line 180: there is no conn argument in the method

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
```diff
@@ -0,0 +1,124 @@
...
+        super().__init__(
+            username,
+            pwd,
+            account=account,
+            warehouse=warehouse,
+            schema=schema,
+            database=database,
+        )
+
+    def connect(self):
```
johnathan-rodriguez, line 38: Not sure if it is worth refactoring this to return 1 or 0 to match the other connect method. Also looked at the current authtest- we may want to log the error message there as well so that the end user gets more details

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+
+        self.logger.info("Attempting to write pandas dataframe to Snowflake")
+        try:
+            session.write_pandas(
+                df, table_name=table_name, auto_create_table=True, overwrite=overwrite
+            )
+            self.logger.info(f"Successfully wrote data to {table_name}")
+
+        except Exception as e:
```
johnathan-rodriguez, line 80: At a glance this error may lead to a misleading exit code. For example if the user attempts to call this method with invalid credentials would that error be raised before this exception is caught? It may say invalid credentials with an exit code `EXIT_CODE_INVALID_UPLOAD_VALUE`

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def put(self, session: Session, file_path: str, table_name: str, overwrite=False):
+        stage_name = f'@%"{table_name}"'
+        if overwrite:
+            try:
+                session.sql(f"CREATE OR REPLACE stage {table_name}").collect()
+                self.logger.info("Successfully created table in Snowflake")
+            except Exception as e:
+                self.logger.error("Error in creating table")
+                raise ExitCodeException(
```
johnathan-rodriguez, line 94: Same question re: error handling here

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
```diff
@@ -0,0 +1,124 @@
...
+                    message=e, exit_code=self.EXIT_CODE_INVALID_UPLOAD_VALUE
+                )
+        self.logger.info("Attempting to put file to Snowflake")
+        try:
+            session.file.put(file_path, stage_name, overwrite=overwrite)
+            self.logger.info("Successfully loaded file into Snowflake")
+        except Exception as e:
```
johnathan-rodriguez, line 101: Same question re: error handling here

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
```diff
@@ -0,0 +1,124 @@
...
+    def copy_into(self, session: Session, table_name: str, overwrite=False):
+        stage_name = f"@%{table_name}"
+        self.logger.info("Attempting to copy into Snowflake")
+        try:
+            copy_statement = f"""COPY INTO "{table_name}" FROM '@%\"{table_name}\"' PURGE=TRUE FILE_FORMAT=(TYPE=CSV FIELD_DELIMITER=',' COMPRESSION=GZIP, PARSE_HEADER=TRUE) MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE"""
+            session.sql(copy_statement)
+            self.logger.info("Successfully copied into Snowflake")
+        except Exception as e:
+            self.logger.error("Error in copying into Snowflake")
+            raise ExitCodeException(
```
johnathan-rodriguez, line 116: Same question re: error handling here

shipyard_blueprints/snowflake/shipyard_snowflake/test/snowflake_test.py
```diff
@@ -0,0 +1,68 @@
...
+import pandas as pd
+from shipyard_snowflake import SnowflakeClient, read_file
+from dotenv import load_dotenv, find_dotenv
+from shipyard_snowflake.utils import (
+    map_pandas_to_snowflake,
+    infer_schema,
+    reservoir_sample,
+)
+
+load_dotenv(find_dotenv())
```
johnathan-rodriguez, line 11: Should add a condition to skip tests if a .env file doesn't exist

shipyard_blueprints/snowflake/shipyard_snowflake/test/snowpark_test.py
```diff
@@ -0,0 +1,63 @@
+import os
+import pandas as pd
+from shipyard_snowflake import SnowparkClient
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
```
johnathan-rodriguez, line 6: add a condition to skip test if .env doesn't exist

shipyard_blueprints/snowflake/shipyard_snowflake/utils.py
```diff
@@ -109,13 +113,13 @@ def map_snowflake_to_pandas(snowflake_data_types: List[List]) -> Union[Dict, Non
             converted = snowflake_to_pandas[str(dtype).upper()]
             if converted is None:
                 raise Exception(
-                    f"Invalid datatypes: the datatype {field} is not a recognized snowflake datatype"
+                    f"Invalid datatypes: the datatype {dtype} is not a recognized snowflake datatype"
                 )
 
             pandas_dtypes[field] = converted
         except KeyError as e:
             raise Exception(
```
johnathan-rodriguez, line 121: should log the error message as well the error message is useful otherwise can clean remove `as e` since the variable is not being used

shipyard_blueprints/snowflake/tests/test_auth.py
```diff
@@ -1,11 +1,13 @@
 import os
 from shipyard_snowflake import SnowflakeClient
 from dotenv import load_dotenv, find_dotenv
+from shipyard_snowflake import SnowparkClient
+from typing import Union
 
 load_dotenv(find_dotenv())
```
johnathan-rodriguez, line 7: Add condition to skip tests that use variables from the env file it it does not exist

shipyard_blueprints/snowflake/tests/test_auth.py
```diff
@@ -18,41 +20,54 @@ def conn_helper(client: SnowflakeClient) -> int:
         return 1
 
 
-def test_good_connection():
```
johnathan-rodriguez, line 21: adding the above condition will make it so you don't have to have code that has been commented out

johnathan-rodriguez (commented):
Minor changes mainly around formatting and questions about the error handling

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
wrp801: For database's and cloud storage, I think it is best if the `connect` method actually returns the connected instance, that is how something like `SQLAlchemy` does it. I think this makes more sense when you need to establish some sort of connection, rather than just call an API

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
wrp801: You would need to actually connect first before running this, so credential errors are actually not possible as those would be caught in the `connect` method

shipyard_blueprints/snowflake/shipyard_snowflake/utils.py
wrp801: Agreed, but the function itself should not log the message, but any class method that calls the function should.

shipyard_blueprints/snowflake/shipyard_snowflake/snowpark.py
johnathan-rodriguez: the invalid credentials was an example of another error that could occur. Ideally the broad exception catch should be UNKNOWN ERROR. If there isn't a more specific error code/msg or exception thrown it may not be possible and invalid upload may still be the best option.

in a perfect world the flow would be more along the lines of
```
try:
    code
except ValueError as error:
     raise .. EXIT_CODE_INVALID_UPLOAD_VALUE
 except exception as error
     raise UNKNOWN ERROR
```
this comment is food for thought

shipyard_blueprints/snowflake/tests/test_auth.py
johnathan-rodriguez: the line before test that need valid env file add
@pytest.mark.skipif(not env_exists, reason="No .env file found")

johnathan-rodriguez (commented):
only thing left is to get the build jobs passing

johnathan-rodriguez (approved):

PR #98, opened 2023-11-15
Notion authtest fix

wrp801:
Refactored `cli/authtest.py`  to not import package but to instead use `requests` to query the `me` endpoint

### DIFF
--- a/shipyard_blueprints/notion/shipyard_notion/cli/authtest.py
+++ b/shipyard_blueprints/notion/shipyard_notion/cli/authtest.py
@@ -1,17 +1,19 @@
 import os
 import sys
-from shipyard_notion import NotionClient
+import requests
 
 
 def main():
-    try:
-        client = NotionClient(token=os.getenv("NOTION_ACCESS_TOKEN"))
-        c = client.connect()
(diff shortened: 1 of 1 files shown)


shipyard_blueprints/notion/shipyard_notion/cli/authtest.py
```diff
@@ -1,17 +1,24 @@
 import os
 import sys
-from shipyard_notion import NotionClient
+import requests
 
 
 def main():
-    try:
-        client = NotionClient(token=os.getenv("NOTION_ACCESS_TOKEN"))
-        c = client.connect()
```
johnathan-rodriguez, line 9: any reason why we the connect function wasn't updated?

wrp801: I just refactored the auth test because the connect method uses the `notion-client` package to establish a client, whereas the auth test simply uses requests. I chose this to not have to install `notion-client` for the root package [](edited-with-Axolo)

johnathan-rodriguez (approved):

PR #106, opened 2023-11-29
Fix Bugs discovered in backwards compatibility tests

wrp801:


shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
```diff
@@ -191,20 +193,36 @@ def get_folder_id(
...
+                )
+
             if folders:
                 return folders[0]["id"]
             else:
                 return None
-    except Exception:
+
+    except ExitCodeException as ec:
+        raise (ExitCodeException(ec.message, ec.exit_code))
```
johnathan-rodriguez, line 224: Not preventing me from approving but a small stylistic suggestion. remove the extra `(` `)`

shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
```diff
@@ -219,7 +237,7 @@ def get_drive_id(drive_id: str, service) -> Union[str, None]:
 
     """
     try:
-        if len(drive_id) == 33 and str(drive_id).startswith("OAJ"):
+        if len(drive_id) == 19 and str(drive_id).startswith("0A"):
```
johnathan-rodriguez, line 240: curious why this condition changed? It would be worth considering making a helper function in the drive utils such as `is_drive_{type}` so that if the criteria needs to be changed again in the future it can be done in 1 place or at the very least add a comment of why char len 19 that begin with 0A is handled differently

shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
```diff
@@ -269,15 +287,146 @@ def get_file_matches(
...
-    folder_id = get_folder_id(folder_id, service=service)
     try:
-        query = f"name contains '{pattern}'"
+        # query = f"name contains '{pattern}'"
+        query = None
         if folder_id:
-            query += f" and '{folder_id}' in parents"
-        results = service.files().list(q=query).execute()
-        files = results.get("files", [])
+            # query += f" and '{folder_id}' in parents"
```
johnathan-rodriguez, line 294: I am going to butcher explaining this but you can clean up the code by setting the query based on if the folder id is provided. that should make one of the else blocks redundant. I would also test if passing None into the functions yeilds the same results. for example if
```.list(
                            q=query,
                            supportsAllDrives=True,
                            includeItemsFromAllDrives=True,
                            corpora="drive",
                            driveId=drive_id,
                        )
```
the same as            
```.list(
                            q=query,
                            supportsAllDrives=True,
                            includeItemsFromAllDrives=True,
                            corpora="drive",
                            driveId=None,
                        )
```

if that is the case you could clean up this code

shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
```diff
@@ -269,15 +287,146 @@ def get_file_matches(
...
+                matches.append(f)
+                id_set.add(f["id"])
     except Exception as e:
         raise ExitCodeException(f"Error in finding matching files: {str(e)}", 210)
 
     else:
-        return results.get("files", [])
+        return matches
+
+    # if folder_id:
```
johnathan-rodriguez, line 386: Do we want to remove this commented out section or did you leave it in on purpose?

shipyard_blueprints/googledrive/shipyard_googledrive/googledrive.py
```diff
@@ -92,24 +92,28 @@ def upload(
...
+                if not self.folder_id and not drive_utils.is_folder_id(
+                    self.folder_name
+                ):
+                    folder_results = drive_utils.create_remote_folder(
+                        folder_name=self.folder_name,
+                        service=self.service,
+                        drive_id=self.drive_id,
+                    )
+                    self.folder_id = folder_results
+                # else:
```
johnathan-rodriguez, line 113: More commented out code

shipyard_blueprints/googledrive/shipyard_googledrive/googledrive.py
```diff
@@ -235,12 +248,12 @@ def download(
             destination_path = os.path.join(destination_path, drive_file_name)
 
         try:
-            file_id = drive_utils.get_file_id(
-                file_name=drive_file_name,
-                service=self.service,
-                drive_id=self.drive_id,
-                folder_id=self.folder_id,
-            )
+            # file_id = drive_utils.get_file_id(
```
johnathan-rodriguez, line 251: More commented out code

johnathan-rodriguez (changes requested):
The only real change I would recommend is clean up `get_file_matches` function and remove the commented out code.

shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
wrp801: This was actually incorrect to begin with. The ID for a folder is 33 chars long, while for a drive it is 19. And the prefix for drives is actually just `0A...` not `OA....`

shipyard_blueprints/googledrive/shipyard_googledrive/drive_utils.py
wrp801: The issue here is that the other parameters like supportsAllDrives, corpora, includeItemsFromAllDrives are only necessary if the `drive_id` is provided, so that's why I had to handle all of those separately

PR #110, opened 2023-12-08
Simplified auth test and replace \n with \n which was throwing an err…

johnathan-rodriguez:
…or when attempting to json.loads()

### DIFF
--- a/shipyard_blueprints/airtable/shipyard_airtable/airtable.py
+++ b/shipyard_blueprints/airtable/shipyard_airtable/airtable.py
@@ -1,6 +1,5 @@
 import requests
 from shipyard_templates import Spreadsheets
-from pyairtable import Table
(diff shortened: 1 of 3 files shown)


johnathan-rodriguez:
when we move the googlesheets to the mono repo it should be able to be simplified to

```
import json

from google.oauth2 import service_account
from shipyard_templates import Spreadsheets


class GoogleSheetsClient(Spreadsheets):
    def __init__(self, service_account: str) -> None:
        self.service_account = service_account
        super().__init__(service_account=service_account)

    def connect(self):
        service_account_str = self.service_account.replace("\n", "\\n")
        service_account.Credentials.from_service_account_info(json.loads(service_account_str))
```

wrp801 (approved):

PR #114, opened 2023-12-12
Bug fix/google auth testv2

johnathan-rodriguez:


shipyard_blueprints/googledrive/shipyard_googledrive/googledrive.py
```diff
@@ -1,13 +1,15 @@
...
+
 from googleapiclient.discovery import build
 from googleapiclient.http import MediaFileUpload
 from google.oauth2 import service_account
 from shipyard_templates import CloudStorage, ExitCodeException
 from typing import Optional, Dict, List, Any, Union
 from googleapiclient.http import MediaIoBaseDownload
-from shipyard_googledrive import drive_utils
+
+from shipyard_blueprints.googledrive.shipyard_googledrive import drive_utils
```
wrp801, line 12: I don't think we want that full path import (`from shipyard_blueprints.googledrive.shipyard_googledrive import ...` ) in place. That may work if you're running it from the root, but is a stand alone package that is not run from the root. So I were to do a version change, build, and then publish this, we would receive an import error not recognizing `shipyard_blueprints`.

shipyard_blueprints/googledrive/shipyard_googledrive/__init__.py
```diff
@@ -1 +1,2 @@
 from .googledrive import GoogleDriveClient
+from shipyard_blueprints.googledrive.shipyard_googledrive import drive_utils
```
wrp801, line 2: Same comment as below regarding the full path.

wrp801 (changes requested):

PR #117, opened 2023-12-19
Microsoft Power BI: Trigger sync blueprint code

johnathan-rodriguez:


shipyard_blueprints/microsoft-power-bi/shipyard_microsoft_power_bi/microsoft_power_bi.py
```diff
@@ -0,0 +1,280 @@
...
+        resource: https://docs.microsoft.com/en-us/rest/api/power-bi/datasets/getdatasetrefreshes
+
+        @param group_id: Group/Workspace ID
+        @param dataset_id: Dataset ID
+        @param request_id: Request ID to check for. This is the ID returned by the refresh request.
+        @param number_of_refreshes: The number of recent refreshes to check.
+        @return: Response from the request.
+        """
+        response = self._request(
+            f"{BASE_URL}/groups/{group_id}/datasets/{dataset_id}/refreshes?$top={number_of_refreshes}",
```
johnathan-rodriguez, line 218: I've considered using $top=1 to fetch the latest run, similar to how we handle the dataflow check.

However, I have some reservations about this method. Primarily, it introduces a higher susceptibility to external factors, such as scheduled refreshes or interventions by other users. By relying on `$top=1`, we might inadvertently capture a run that wasn't initiated by our API call. This could lead to unintended consequences or confusion.  I didn't see the option to do this with dataflows but dataflows do not allow for multiple refreshes to be queued.

blakeburch:
Is there no way to get an ID of the initial refresh to ensure you're checking the right one?

johnathan-rodriguez:
for `dataset` we can grab it after the request is made because the request id is part of the refreshes endpoint`s response which is not the case for `dataflows` . For this reason I am confident the status we're grabbing from the `datasets` refresh is indeed the one triggered by our blueprint

That is not the case with the Dataflows Transactions endpoint. We could be more confident on Dataflows based on the start time of the transaction but since I didn't find any documentation that clearly stated how microsoft is handling the start time I opted for the latest run

shipyard_blueprints/microsoft-power-bi/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "shipyard-microsoft-power-bi"
+version = "0.1.0"
```
wrp801, line 3: Go ahead and make this pre-release too

shipyard_blueprints/microsoft-power-bi/shipyard_microsoft_power_bi/cli/trigger_sync.py
```diff
@@ -0,0 +1,54 @@
...
+
+        client.refresh(
+            object_type=args.object_type,
+            object_id=args.object_id,
+            group_id=args.group_id,
+            wait_for_completion=wait_for_completion,
+            wait_time=int(args.poke_interval),
+        )
+    except ExitCodeException as e:
+        print(e)
```
wrp801, line 50: These print statements should be logged instead

shipyard_blueprints/microsoft-power-bi/shipyard_microsoft_power_bi/cli/trigger_sync.py
```diff
@@ -0,0 +1,54 @@
...
+    parser.add_argument("--client-id", dest="client_id", required=True)
+    parser.add_argument("--client-secret", dest="client_secret", required=True)
+    parser.add_argument("--tenant-id", dest="tenant_id", required=True)
+    parser.add_argument("--group-id", dest="group_id", required=True)
+    parser.add_argument("--object-type", dest="object_type", required=True)
+    parser.add_argument("--object-id", dest="object_id", required=True)
+    parser.add_argument(
+        "--wait-for-completion", dest="wait_for_completion", required=True
+    )
+    parser.add_argument("--poke-interval", dest="poke_interval", required=False)
```
wrp801, line 18: Set a default value of this for 60 seconds, since that is what we will do in the application

shipyard_blueprints/microsoft-power-bi/shipyard_microsoft_power_bi/microsoft_power_bi_utils.py
```diff
@@ -0,0 +1,201 @@
...
+        client_id=bi_instance.client_id,
+        authority=f"https://login.microsoftonline.com/{bi_instance.tenant_id}",
+    ).acquire_token_by_username_password(
+        username=bi_instance.username,
+        password=bi_instance.password,
+        scopes=["https://analysis.windows.net/powerbi/api/.default"],
+    )
+
+
+class PowerBiUtils:
```
wrp801, line 33: I'm not going to get hung up on it, but given that all the methods in here are static methods, it probably didn't need to be a class. Could have just be a set of functions

wrp801 (changes requested):
Small changes to make

wrp801 (approved):

PR #119, opened 2023-12-20
Added the utils to the init and added logging where needed

johnathan-rodriguez:


shipyard_blueprints/microsoft-power-bi/shipyard_microsoft_power_bi/microsoft_power_bi.py
```diff
@@ -2,7 +2,7 @@
 
 from requests import request
 from shipyard_templates import DataVisualization, ExitCodeException
-import microsoft_power_bi_utils as PowerBiUtils
+from shipyard_microsoft_power_bi import microsoft_power_bi_utils as bi_utils
```
johnathan-rodriguez, line 5: suggestion on a better alias is welcomed. Changed from PowerBiUtils since it is no longer a class

wrp801 (approved):

PR #129, opened 2024-01-11
Removed the need for the legacy folder for the consolidation

johnathan-rodriguez:


shipyard_blueprints/email/shipyard_email/email.py
```diff
@@ -70,174 +73,3 @@ def connect(self):
         else:
             self.logger.error("Signin method provided was not tls or ssl")
             return 1
-
```
johnathan-rodriguez, line 73: Removed to avoid confusion when we do refactor project

wrp801 (approved):

PR #134, opened 2024-01-17
Jr/integ 347 shipyard templates logger

johnathan-rodriguez:


johnathan-rodriguez:
auth tests would something like

```
import os
import sys

from shipyard_microsoft_power_bi import MicrosoftPowerBiClient
from shipyard_templates.shipyard_logger import ShipyardLogger


def main():
    logger = ShipyardLogger.get_logger()
    logger.setLevel(ShipyardLogger.AUTHTEST_LEVEL)
    sys.exit(
        MicrosoftPowerBiClient(
            client_id=os.getenv("MICROSOFT_POWER_BI_CLIENT_ID"),
            client_secret=os.getenv("MICROSOFT_POWER_BI_CLIENT_SECRET"),
            tenant_id=os.getenv("MICROSOFT_POWER_BI_TENANT_ID"),
        ).connect()
    )


if __name__ == "__main__":
    main()
```

class files will look like
```
from shipyard_templates import DataVisualization, ExitCodeException,ShipyardLogger
logger = ShipyardLogger.get_logger()

class MicrosoftPowerBiClient(DataVisualization):
    def __init__(self, client_id: str, client_secret: str, tenant_id: str, **kwargs):
        """
        Initializes the MicrosoftPowerBiClient with the provided credentials.

        Args:
            client_id (str): Client ID for Power BI API authentication.
            client_secret (str): Client Secret for Power BI API authentication.
            tenant_id (str): Tenant ID for Power BI API authentication.
            **kwargs: Additional keyword arguments passed to the superclass.
        """

        super().__init__(**kwargs)
        self.access_token = None
        self.client_id = client_id
        self.client_secret = client_secret
        self.tenant_id = tenant_id

        logger.info("MicrosoftPowerBiClient Initialized")
    def connect(self):
        """
        Confirms that the client can connect to the Power BI API.

        @return: 1 if the connection fails, 0 otherwise.
        """
        try:
            bi_utils.generate_access_token(self)

            logger.error(e)
            # Custom messaging here
            logger.authtest(f"Connection failed")
            return 1
        else:
            return 0
```

shipyard-templates/shipyard_templates/shipyard_logger.py
```diff
@@ -1,18 +1,46 @@
 import logging
 
 
+def addLoggingLevel(levelName, levelNum, methodName=None):
+    """
```
wrp801, line 5: Small change, but these functions and nested functions should be snake case, not camel case

wrp801 (changes requested):
Small stylistic change in the `shipyard_logger.py` module

wrp801 (approved):

PR #138, opened 2024-01-26
Bumped version and included log level to be pulled from env var

johnathan-rodriguez:


shipyard-templates/shipyard_templates/shipyard_logger.py
```diff
@@ -34,6 +35,8 @@ class ShipyardLogger:
     def get_logger(cls):
         if cls._logger is None:
             cls._logger = logging.getLogger("Shipyard")
+            log_level = os.getenv("LOG_LEVEL", "INFO").upper()
```
johnathan-rodriguez, line 38: LOG_LEVEL will be the env key when using run with params

wrp801 (approved):

PR #143, opened 2024-02-02
Jr/integ 24 airtable

johnathan-rodriguez:


shipyard_blueprints/airtable/pyproject.toml
```diff
@@ -1,21 +1,20 @@
...
 license = "Apache 2.0"
 readme = "README.md"
-packages = [{include = "shipyard_airtable"}]
+packages = [{ include = "shipyard_airtable" }]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-pyairtable = "^1.4.0"
+pyairtable = "2.2.2"
 pandas = "^1.5.3"
```
wrp801, line 13: bump pandas to at least 2.0, you could try doing the latest version too

shipyard_blueprints/airtable/shipyard_airtable/airtable.py
```diff
@@ -1,24 +1,113 @@
...
+    EXIT_CODE_INVALID_BASE = 101
+    EXIT_CODE_INVALID_TABLE = 102
+    EXIT_CODE_INVALID_VIEW = 103
+    EXIT_CODE_RESOURCE_NOT_FOUND = 104
+
+    def __init__(self, api_key: str) -> None:
+        self.api = Api(api_key)
+
+    def _handle_error(self, error: HTTPError):
+        logger.debug("Handling HTTPError...")
```
wrp801, line 19: Add docstrings to all the methods, including the private ones

shipyard_blueprints/airtable/shipyard_airtable/cli/upload.py
```diff
@@ -0,0 +1,118 @@
+import sys
+import pandas
```
wrp801, line 2: This needs to include one additional argument for `Insert Method`, which is a choice of `replace` or `append`. Right now I'm assuming that it just overwrites the existing data?

wrp801 (changes requested):
In summary, add docstrings to all the methods in the class as well as the helpers in the CLI, and add an `Insert Method` parameter for the upload cli.

shipyard_blueprints/airtable/shipyard_airtable/cli/upload.py
johnathan-rodriguez: For clarification this is currently doing an upsert. 

Do we still want to add the replace or append. The end user will able to accomplish replace or append by omitting or including the record id unless replace is intended to clear out the table entirely so that the only data that exists what is in the csv file upload

wrp801: Yes - to match functionality of our other uploads we should add in the ability to overwrite an existing table with the data being uploaded. [](edited-with-Axolo)

johnathan-rodriguez: I would argue that upsert is a useful and intuitive function, should I include it as an option considering not all services may not have an endpoint that supports that out of the box?

wrp801: Yeah thats fine [](edited-with-Axolo)

shipyard_blueprints/shipyard-bp-utils/shipyard_bp_utils/files.py
```diff
@@ -260,8 +262,8 @@ def find_all_file_matches(file_names: list, file_name_re: str) -> list:
     """
     matching_file_names = [file for file in file_names if re.search(file_name_re, file)]
 
-    logger.info(f"Found {len(matching_file_names)} file matches.")
```
wrp801, line 263: I think this line that prints out the file matches should be left in the `info` level. That is useful information to the users that should be available outside of `debug` mode

wrp801 (changes requested):
Small change on the bp-utils package, I think one of the log statements should be left at the info level

wrp801 (approved):

PR #149, opened 2024-02-06
Added support for third insertion method

wrp801:


shipyard_blueprints/snowflake/shipyard_snowflake/cli/execute_sql.py
```diff
@@ -55,6 +55,7 @@ def main():
         sys.exit(e.exit_code)
     else:
         logger.info("Successfully executed query")
+        logger.info("Closing Connection")
```
johnathan-rodriguez, line 58: should be combined into 1 log line since no other code is being executed in between. If you want the new line for the styling should use \n

shipyard_blueprints/snowflake/shipyard_snowflake/cli/upload.py
```diff
@@ -53,7 +53,7 @@ def get_args():
     parser.add_argument(
         "--insert-method",
         dest="insert_method",
-        choices={"fail", "replace", "append"},
+        choices={"add", "replace", "append"},
```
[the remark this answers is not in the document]
wrp801: I changed the internal value in the blueprint and already did a regression test, so we should be good

shipyard_blueprints/snowflake/shipyard_snowflake/cli/upload.py
```diff
@@ -164,8 +164,8 @@ def main():
             # upload each file to the target
             # NOTE: This should be reworked to PUT multiple files at once using a glob pattern, and COPY INTO multiple
             # files at once using the FILES() parameter
+            insert_method = args.insert_method
             for i, file_match in enumerate(matching_file_names, start=1):
```
johnathan-rodriguez, line 168: i believe we have a utils for this. Could it be used here?

shipyard_blueprints/snowflake/shipyard_snowflake/cli/upload.py
```diff
@@ -217,6 +217,9 @@ def main():
             f"An error occurred when attempting to upload to Snowflake: {str(e)}"
         )
         sys.exit(Database.EXIT_CODE_UNKNOWN)
+    else:
+        logger.info("Closing connection")
+        snowflake_client.close()
```
johnathan-rodriguez, line 222: snowflake_client.close() is a precaution but should be moved into `finally` so that it always executes

shipyard_blueprints/snowflake/shipyard_snowflake/snowflake.py
```diff
@@ -118,13 +119,18 @@ def upload(
             table_name (str): The name of the Snowflake Table to, if it doesn't exist, it will be created
             insert_method (str): The method to use when inserting the data into the table. Options are replace or append Defaults to 'replace'
         """
-        if insert_method not in ["replace", "append"]:
+        if insert_method not in ["replace", "append", "add"]:
```
johnathan-rodriguez, line 122: not worth the change but worth noting ideally we should use a set for these types of checks. Slightly faster.
{"replace", "append", "add"}:

shipyard_blueprints/snowflake/shipyard_snowflake/cli/upload.py
wrp801: I had it in the finally initially but put it here because in the finally clause, there is a chance the client hasn't been initialized so it would throw an exception

shipyard_blueprints/snowflake/shipyard_snowflake/utils/exceptions.py
```diff
@@ -56,3 +56,10 @@ def __init__(self, message: str, exit_code: int):
         super().__init__(message, exit_code)
         self.message = message
         self.exit_code = exit_code
+
+
+class TableRecordsError(ExitCodeException):
```
johnathan-rodriguez, line 61: I don't see this being used anywhere

johnathan-rodriguez (approved):

PR #153, opened 2024-02-08
Adjusted error handling when error details are not nested and added a…

johnathan-rodriguez:
… catch-all in case there are any other inconsistencies

### DIFF
--- a/shipyard_blueprints/airtable/shipyard_airtable/airtable.py
+++ b/shipyard_blueprints/airtable/shipyard_airtable/airtable.py
@@ -28,57 +28,73 @@ class AirtableClient(Spreadsheets):
         logger.debug(
             f"Airtable Server Status Code: {error.response.status_code} \n Server Reason: {error.response.reason}"
         )
-        response_details = error.response.json()
-        error_details = response_details.get("error", {})
 
-        if error.response.status_code == 401:
+        try:
+            response_details = error.response.json()
+
+            error_details = response_details.get("error", {})
+            if isinstance(error_details, str):
+                error_details = {"type": error_details}
+
+            if error.response.status_code == 401:
+                raise ExitCodeException(
+                    error_details.get(
+                        "message",
+                        "The API key provided is incorrect."
+                        "Please double-check for extra spaces or invalid characters.",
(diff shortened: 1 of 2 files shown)


shipyard_blueprints/airtable/shipyard_airtable/airtable.py
```diff
@@ -28,57 +28,73 @@ def _handle_error(self, error: HTTPError) -> None:
...
-                    "message", "The rate limit for the API key has been exceeded."
-                ),
-                self.EXIT_CODE_RATE_LIMIT,
-            )
-        else:
+        try:
+            response_details = error.response.json()
+
+            error_details = response_details.get("error", {})
+            if isinstance(error_details, str):
```
johnathan-rodriguez, line 36: There are scenarios where the server returns {"error":"error_reason"} instead of the more common 
```{'error': {'message': 'Invalid permissions, or the requested model was not found. Check that both your user and your token have the required permissions, and that the model names and/or ids are correct.', 'type': 'INVALID_PERMISSIONS_OR_MODEL_NOT_FOUND'}}```

This was causing a `'str' object has no attribute 'get'`

File: shipyard_blueprints/airtable/shipyard_airtable/airtable.py [hunk 2]
```diff
@@ -28,57 +28,73 @@ def _handle_error(self, error: HTTPError) -> None:
         logger.debug(
```
wrp801, line 28: Throw this debug message in the CLI portion and remove it from here

wrp801 (changes requested):

shipyard_blueprints/airtable/shipyard_airtable/airtable.py
johnathan-rodriguez: I would push a bit on this because it is necessary debugging information specific to the `_handle_error`. 

what I was referring to in slack was moving the message from the catch all exception into the cli portion but after thinking it through a bit more it would be useful in the class to avoid duplicating code between blueprints.

wrp801 (approved):

PR #154, opened 2024-02-08
Wes/integ 370 bugfix download with folder error

wrp801:
Fixing bug in download cli where the target folder is not being created properly


johnathan-rodriguez (approved):
Approved with caveat-

consider using `shipyard-bp-utils.files.create_folder_if_dne` for consistency or for simplicity consider `os.mkdir(destination_folder_name,exist_ok=True)`

PR #156, opened 2024-02-12
Wes/integ 262 google bigquery

wrp801:


shipyard_blueprints/bigquery/shipyard_bigquery/utils/exceptions.py
```diff
@@ -0,0 +1,50 @@
+from shipyard_templates import ExitCodeException
+
+
+class DatasetNotFound(ExitCodeException):
+    def __init__(self, message: str, exit_code: int):
+        super().__init__(message, exit_code)
+        self.message = message
+        self.exit_code = exit_code
```
johnathan-rodriguez, line 8: I would suggest with the custom exceptions to include the exit code and possibly the message as well.

shipyard_blueprints/bigquery/shipyard_bigquery/bigquery.py
```diff
@@ -1,39 +1,225 @@
+import json
```
johnathan-rodriguez, line 1: I believe this code change would be picked up by the prod blueprints. You may have to move the class code changes under the shipyard folder to avoid unintentional commits when merged into main

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    try:
+        args = get_args()
+        target_folder = (
+            args.destination_folder_name if args.destination_folder_name != "" else None
```
johnathan-rodriguez, line 38: can be simplified to 
```
if target_folder := args.destination_folder_name:
...
```

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download.py
```diff
@@ -0,0 +1,69 @@
...
+        logger.debug(f"Shape of the data is {df.shape}")
+        df.to_csv(target_path, index=False)
+    except ExitCodeException as ec:
+        logger.error(ec.message)
+        sys.exit(ec.exit_code)
+    except Exception as e:
+        logger.error(
+            f"Error in fetching query results and writing results to {target_path}: {str(e)}"
+        )
+        sys.exit(client.EXIT_CODE_NO_RESULTS)
```
johnathan-rodriguez, line 62: This will throw an error in scenarios where the client init fails

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download_to_gcs.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    try:
+        args = get_args()
+        destination_file_name = args.destination_file_name
+        destination_folder_name = (
```
johnathan-rodriguez, line 37: Similar simplification here as well

if empty_str returns False

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download_to_gcs.py
```diff
@@ -0,0 +1,67 @@
...
+
+    except exceptions.DownloadToGcsError as de:
+        logger.error(de.message)
+        sys.exit(de.exit_code)
+    except ExitCodeException as ec:
+        logger.error(ec.message)
+        sys.exit(ec.exit_code)
+    except Exception as e:
+        logger.error(f"An error occurred in downloading query results to GCS: {str(e)}")
+        sys.exit(client.EXIT_CODE_NO_RESULTS)
```
johnathan-rodriguez, line 61: Same here. client.EXIT_CODE will throw another error if client doesnt init

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/execute_query.py
```diff
@@ -0,0 +1,42 @@
...
+        client.execute_query(query)
+    except QueryError as qe:
+        logger.error(qe.message)
+        sys.exit(qe.exit_code)
+    except ExitCodeException as ec:
+        logger.error(ec.message)
+        sys.exit(ec.exit_code)
+    except Exception as e:
+        logger.error(f"Error in executing query: {str(e)}")
+        sys.exit(client.EXIT_CODE_QUERY_ERROR)
```
johnathan-rodriguez, line 36: Same here.

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/upload.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+
+            for index, file_name in enumerate(matching_file_names, start=1):
+                logger.info(f"Uploading file {index} of {len(matching_file_names)}")
+                client.upload(
+                    file=file_name,
+                    dataset=dataset,
+                    table=table,
+                    upload_type=upload_type,
+                    skip_header_rows=skip_header_rows,
```
johnathan-rodriguez, line 97: Does this work as expected if `skip_header_rows=""` if so fine as is otherwise i recommend
`if args.skip_header_rows:
     skip_header_rows=int(args.skip_header_rows)
  else:
  logger.warning("skip header row was not defined. Continuing with default value of 1")
  ```

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/upload.py
```diff
@@ -0,0 +1,131 @@
...
+        sys.exit(ie.exit_code)
+    except SchemaFormatError as se:
+        logger.error(se.message)
+        sys.exit(se.exit_code)
+    except ExitCodeException as ec:
+        logger.error(ec.message)
+        sys.exit(ec.exit_code)
+    except Exception as e:
+        logger.error(f"Error in uploading file to BigQuery: {str(e)}")
+        sys.exit(client.EXIT_CODE_INVALID_UPLOAD_VALUE)
```
johnathan-rodriguez, line 127: same issue as before here as well

johnathan-rodriguez (changes requested):

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download.py
wrp801: going to leave the walrus operator out of it this time

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/download_to_gcs.py
wrp801: What is the simplification here?

shipyard_blueprints/bigquery/shipyard_bigquery/shipyard/upload.py
wrp801: This does work because it is set to `None` above, and the method defined in the class defaults to None

johnathan-rodriguez:
As promised a better indented example along with a clearer explanation

for areas where you have `variable = args.value if args.value !="" else None` and then `None` is only used in a `if variable:` condition at its core you don't really care if the arg value is an `""` but more so if the expression of `args.value` is `True` or `False`. 

An example of where this is more apparent is the following section:
```
        target_folder = (
            args.destination_folder_name if args.destination_folder_name != "" else None
        )
        if target_folder: # None here is only used to branch off. If taget_folder="" this condition would still be met
            shipyard.files.create_folder_if_dne(target_folder)

            target_path = shipyard.files.combine_folder_and_file_name(
                folder_name=target_folder, file_name=args.destination_file_name
            )
        else:
            target_path = os.path.join(os.getcwd(), args.destination_file_name)
```

Since the following is not a valid option
```
target_path = shipyard.files.combine_folder_and_file_name(folder_name=None, file_name=args.destination_file_name)
```                            
`target_folder = (args.destination_folder_name if args.destination_folder_name != "" else None)`
is not needed. Rather the intention of the core more aligns with

```
target_folder = args.destination_folder_name or os.getcwd()
shipyard.files.create_folder_if_dne(target_folder)
target_path = shipyard.files.combine_folder_and_file_name(folder_name=target_folder,
                                                          file_name=args.destination_file_name)
```

This remove the need for an `if` statement handling `None` especially in scenarios where the `None` path is not a viable option. Takes a 9 line code and reduces it to 4 where the codes intention is clear which makes maintaining the code much easier as well.

shipyard_blueprints/bigquery/shipyard_bigquery/utils/exceptions.py
```diff
@@ -0,0 +1,50 @@
...
+EXIT_CODE_DOWNLOAD_TO_GCS_ERROR = 103
+EXIT_CODE_TEMP_TABLE_CREATION_ERROR = 104
+EXIT_CODE_SCHEMA_FORMATTING_ERROR = 105
+EXIT_CODE_SCHEMA_VALIDATION_ERROR = 106
+
+
+class InvalidSchema(ExitCodeException):
+    def __init__(self, message: str):
+        self.message = message
+        self.exit_code = 204
```
johnathan-rodriguez, line 14: this would be better as 
from shipyard_template import GoogleDatabase

...
self.exit_code= GoogleDatabase.EXIT_CODE_INVALID_SCHEMA

johnathan-rodriguez (approved):
A minor improvement would be to refer to the base class exit code when raise exit code 204 instead of having it hard coded. Other than that looks good

PR #158, opened 2024-02-15
Included a class function for extracting the a file context into the …

johnathan-rodriguez:
…message.

### DIFF
--- a/shipyard-templates/shipyard_templates/messaging.py
+++ b/shipyard-templates/shipyard_templates/messaging.py
@@ -1,4 +1,7 @@
+import re
 from abc import ABC, abstractmethod
+
+FILE_PATTERN = re.compile(r"\{\{[^\{\}]+\}\}")
 
 
 class Messaging(ABC):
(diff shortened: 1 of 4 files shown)


shipyard-templates/shipyard_templates/messaging.py
```diff
@@ -12,3 +15,64 @@ def connect(self):
     EXIT_CODE_BAD_REQUEST = 203
     EXIT_CODE_FILE_NOT_FOUND = 204
     EXIT_CODE_UNKNOWN_ERROR = 249
+
+    @classmethod
+    def message_content_file_injection(self, message: str) -> str:
```
wrp801, line 20: I'm almost of the opinion that this should be a static method, not a class method, thoughts?

wrp801 (approved):
code is fine, interested to know your thoughts on the static method vs class method

shipyard-templates/shipyard_templates/messaging.py
johnathan-rodriguez: I had the seem thought initially until i wanted to raise the exit code.

PR #163, opened 2024-02-20
Jr/integ 256 rewriterefactor v2

johnathan-rodriguez:


johnathan-rodriguez:
I made a wrapper for handling the exceptions. Would be interested on getting your thoughts on this approach

shipyard_blueprints/email/shipyard_email/__init__.py
```diff
@@ -1 +0,0 @@
-from .email import EmailClient
```
wrp801, line 1: Any reason why this is being removed?

shipyard_blueprints/email/shipyard_email/cli/send_conditional_email.py
```diff
@@ -0,0 +1,160 @@
...
+    try:
+        args = get_args()
+        send_method = args.send_method.lower() or "tls"
+        sender_address = args.sender_address
+        username = args.username
+        message = args.message
+        conditional_send = args.conditional_send
+        file_upload = args.file_upload
+        if args.include_shipyard_footer:
+            include_shipyard_footer = shipyard.args.convert_to_boolean(
```
wrp801, line 90: Can we agree to not use the `convert_to_boolean` moving forward. I like the `.strip().upper() == 'TRUE'` approach

shipyard_blueprints/email/shipyard_email/cli/send_email_with_file.py
```diff
@@ -0,0 +1,155 @@
...
+def main():
+    try:
+        args = get_args()
+        send_method = args.send_method.lower() or "tls"
+        sender_address = args.sender_address
+        username = args.username
+        message = args.message
+
+        if args.include_shipyard_footer:
+            include_shipyard_footer = shipyard.args.convert_to_boolean(
```
wrp801, line 91: same comment regarding the `convert_to_boolean` as above

shipyard_blueprints/email/shipyard_email/cli/send_email_with_file.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+
+        source_file_name = args.source_file_name
+        source_folder_name = shipyard.files.clean_folder_name(args.source_folder_name)
+        file_paths = shipyard.files.find_matching_files(
+            source_file_name, source_folder_name, args.source_file_name_match_type
+        )
+
+        if not file_paths:
+            raise InvalidFileInputError(
```
wrp801, line 112: Will this be caught in the `raise ExitCodeException` part? If not, then that needs to be handled and exited

wrp801 (changes requested):
Small code changes and comments. Regarding the exception handling, I think it is fine technically speaking, but I'm not sure how I feel about it. Personally I think it introduces additional complexity. For instance, if I look at a method, I cannot tell what exception it is going to raise if it is decorated with the `@handle_exception` operator. I know it is present in the docstrings, but I prefer to see within the function itself what exception it would raise. I also think that is easier from a maintenance standpoint.

shipyard_blueprints/email/shipyard_email/cli/send_conditional_email.py
johnathan-rodriguez: convert_to_boolean does the `string.upper().strip()` but it is more explicit. I like the thought of being more explicit- i think changing convert_to_boolean not to raise a value error if the string is not true or false but instead return a default value would be the better approach and log a warning so the user is made aware.

```
def convert_to_boolean(string: str, default: bool = False) -> bool:
    """
    Convert a string to a boolean.

    Args:
    string (str): A string to convert to a boolean.

    Returns:
    bool: True if the string is "TRUE" (case-insensitive) else False.
    """
    logger.debug(f"Converting {string} to boolean...")

    bool_string = str(string).upper().strip()  # Handles if string is None
    if bool_string == "TRUE":
        result = True
    elif bool_string == "FALSE":
        result = False
    else:
        result = default
        logger.warning(f'"{string}" is not a valid boolean. Defaulting to {default}.')

    logger.debug(f"Converted {string} to {result} boolean.")
    return result
    ```

File: shipyard_blueprints/email/shipyard_email/cli/send_email_with_file.py (hunk 4 above)
Comment (johnathan-rodriguez) [reply shipyard_blueprints/email/shipyard_email/cli/send_email_with_file.py:112]: it would be caught with `ExitCodeException` it is an exception derived from that class

File: shipyard_blueprints/email/shipyard_email/cli/send_conditional_email.py [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+                args.include_shipyard_footer
+            )
+        else:
+            logger.warning("include_shipyard_footer not set. Defaulting to TRUE.")
+            include_shipyard_footer = True
```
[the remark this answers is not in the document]
johnathan-rodriguez: pasting mainly for a sanity check
image

wrp801 (approved):

wrp801:
approved based off of our conversation

PR #165, opened 2024-02-22
Wes/integ 197 amazon s3

wrp801:


shipyard_blueprints/s3/integration/int.py
```diff
@@ -0,0 +1,245 @@
+import os
```
johnathan-rodriguez, line 1: An particular reason to breakaway from pytest naming convention? Typically, test files are located in a `tests` folder and include `test` as a prefix or suffix in their filenames.

shipyard_blueprints/s3/integration/int.py
```diff
@@ -0,0 +1,245 @@
...
+        "--source-file-name-match-type",
+        "exact_match",
+        "--destination-file-name",
+        "new_s3.csv",
+    ]
+    subprocess.run(download_command, check=True)
+
+    # Check if files match
+    original_rows = sum(1 for _ in open("s3.csv"))
+    new_rows = sum(1 for _ in open("new_s3.csv"))
```
johnathan-rodriguez, line 71: not something that needs to be changed but would also be interesting to check for would be file size. Row count is enough for testing purposes.

shipyard_blueprints/s3/shipyard_s3/cli/authtest.py
```diff
@@ -1,31 +1,26 @@
 import os
 import sys
-import boto3
 from shipyard_s3 import S3Client
+from shipyard_templates import ShipyardLogger
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
```
johnathan-rodriguez, line 7: I understand the need for this running locally but we would probably want to remove this to be on the safe side when it goes to the application

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,173 @@
...
+        try:
+            logger.debug("Establishing connection with S3")
+            client = boto3.client(
+                "s3",
+                region_name=self.region,
+                aws_access_key_id=self.aws_access_key,
+                aws_secret_access_key=self.aws_secret_access_key,
+                config=None,
+            )
+        except Exception as e:
```
johnathan-rodriguez, line 41: may want to include logger.authtest for any useful information we would want the customer to read on a failed connection. This will help customers troubleshoot their credentials issues. This may require handling more specific exceptions raised by s3 depending how readable their errors are

shipyard_blueprints/s3/shipyard_s3/cli/authtest.py
```diff
@@ -1,31 +1,26 @@
...
 def main():
-    s3 = S3Client(
-        aws_access_key=os.getenv("AWS_ACCESS_KEY_ID"),
-        aws_secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
-    )  # only necessary for logging
     try:
-        client = boto3.client(
-            "sts",
-            aws_access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
+        s3 = S3Client(
```
johnathan-rodriguez, line 14: may be worth writing tests for the variations of connect. Mainly thinking of the optional region field. The previous method was not attempting to return the client but return the caller's identity.

shipyard_blueprints/s3/shipyard_s3/cli/download.py
```diff
@@ -43,244 +48,116 @@ def get_args():
...
+        src_file = args.source_file_name
+        src_folder = (
+            shipyard.files.clean_folder_name(args.source_folder_name)
+            if args.source_folder_name
+            else None
+        )
+        match_type = args.source_file_name_match_type
+        target_dir = shipyard.files.clean_folder_name(args.destination_folder_name)
 
+        if not os.path.exists(target_dir) and (target_dir != ""):
```
johnathan-rodriguez, line 86: why not use shipyard.files.create_folder_dne
or directly call `os.makedirs(target_dir, exist_ok=True)`

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,173 @@
...
+            raise RemoveError(message=str(e))
+        else:
+            logger.debug(f"Response from s3: {s3_response}")
 
-    def upload_files(self):
-        pass
+    def upload(
+        self,
+        s3_conn,
```
johnathan-rodriguez, line 98: ` s3_conn` is being used throughout the methods. thoughts on adding it as a class attribute? My thought process is that it would centralize the connection setup, reduce redundancy, and it isn't obvious at a glance that running .connect() is a prerequisite. I also don't see the benefit of a class handling and s3 connection that is not associated with the credentials used to standup the class. What are your thoughts on this?

johnathan-rodriguez (commented):
What is the gameplan with the "old" directory? I assume this will be removed when it is ready for production. Also it could be the lack of caffeine but im not seeing `s3_config` supported in the newer code. Is this intentional?

wrp801:
I moved that so I can run some additional tests - it will be removed

shipyard_blueprints/s3/shipyard_s3/cli/authtest.py
wrp801: Yes this should be removed

shipyard_blueprints/s3/shipyard_s3/s3.py
wrp801: I initially had it like that and then I went away from it and I honestly can't remember why. I think this should be a property since it's reused a bunch

shipyard_blueprints/s3/pyproject.toml
```diff
@@ -1,21 +1,30 @@
...
 
 [tool.poetry.dependencies]
-python = ">=3.7,<3.10"
-boto3 = "1.26"
-shipyard-templates = "0.4.3"
-shipyard-utils = "^0.1.4"
+python = "^3.9"
+boto3 = "1.34.44"
+shipyard-templates = "0.6.2"
+shipyard-bp-utils = "1.0.2a1"
```
johnathan-rodriguez, line 14: this could get bump now

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,177 @@
...
+            copy_source = {"Bucket": src_bucket, "Key": src_path}
+            # move the object(s)
+            self.s3_conn.copy_object(
+                Bucket=dest_bucket, CopySource=copy_source, Key=dest_path
+            )
+            # delete the original
+            self.s3_conn.delete_object(Bucket=src_bucket, Key=src_path)
+
+        except Exception as e:
+            if "IllegalLocationConstraintException" in str(e):
```
johnathan-rodriguez, line 76: this shouldn't be nested.

if should be more like
``` 
from boto3.exceptions import IllegalLocationConstraintException
try:
    ...
except IllegalLocationConstraintException as err:
     raise InvalidRegion(region) from err
except Exception as err:
    raise MoveError('An error occurred when attempting to move file. Message from s3 server: {err})
```

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,177 @@
...
 
-    def remove_files(self):
-        pass
+        Raises:
+            RemoveError:
+        """
+        try:
+            s3_response = self.s3_conn.delete_object(Bucket=bucket_name, Key=src_path)
+        except Exception as e:
+            if "IllegalLocationConstraintException" in str(e):
```
johnathan-rodriguez, line 94: Same comment as the above. Handle the specific exception and not parsing the the string of the error message

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,177 @@
...
+            bucket_name: The name of the bucket to load to
+            source_file: The name or file path of the target file to load
+            destination_path: The optional path of where the file should be loaded to
+            extra_args: Additional optional configuration arguments to be passed
+
+        Raises:
+            UploadError
+        """
+        try:
+            s3_upload_config = boto3.s3.transfer.TransferConfig()
```
johnathan-rodriguez, line 120: Is this needed if s3_config is None?

shipyard_blueprints/s3/shipyard_s3/s3.py
```diff
@@ -1,39 +1,177 @@
...
+            s3_transfer = boto3.s3.transfer.S3Transfer(
+                client=self.s3_conn, config=s3_upload_config
+            )
+            s3_transfer.upload_file(
+                source_file, bucket_name, destination_path, extra_args=extra_args
+            )
+
+        except S3UploadFailedError as se:
+            if "IllegalLocationConstraintException" in str(se):
```
johnathan-rodriguez, line 129: Same comment re: handling the exception

johnathan-rodriguez (changes requested):
Handle the IllegalLocationConstraintException exception

shipyard_blueprints/s3/shipyard_s3/s3.py
wrp801: The problem is that is not an actual exception, it is just present in the message [](edited-with-Axolo)

shipyard_blueprints/shipyard-bp-utils/shipyard_bp_utils/files.py
```diff
@@ -231,12 +232,33 @@ def find_all_local_file_names(source_folder_name: str = None) -> list:
     list: A list of all files that exist in the current working directory.
     """
     logger.debug(f"Finding all local file names in {source_folder_name}...")
-    cwd = os.getcwd()
-    cwd_extension = os.path.normpath(f"{cwd}/{source_folder_name}/**")
-    all_paths = glob.glob(cwd_extension, recursive=True)
-    logger.debug(f"Found {len(all_paths)}.")
-    logger.debug(all_paths)
-    return remove_directories_from_path_list(all_paths)
+    # cwd = os.getcwd()
```
wrp801, line 235: I will remove this commented out code upon approval, just wanted to get your thoughts on the change

shipyard_blueprints/shipyard-bp-utils/shipyard_bp_utils/files.py
```diff
@@ -231,12 +232,33 @@ def find_all_local_file_names(source_folder_name: str = None) -> list:
...
+    # logger.debug(f"Found {len(all_paths)}.")
+    # logger.debug(all_paths)
+    # return remove_directories_from_path_list(all_paths)
+    if source_folder_name is None:
+        source_folder_name = (
+            os.getcwd()
+        )  # Use current working directory if no directory provided
+    try:
+        file_list = [
+            os.path.join(source_folder_name, filename)
```
johnathan-rodriguez, line 251: the original script looks like it intentionally removed the source folder. did we test to make sure this doesn't break `find_matching_files`

johnathan-rodriguez: Also `os.path.join(source_folder_name, filename)` is being used twice here. I recommend using a variable name.

shipyard_blueprints/shipyard-bp-utils/shipyard_bp_utils/files.py
```diff
@@ -231,12 +232,33 @@ def find_all_local_file_names(source_folder_name: str = None) -> list:
...
+    # if not source_folder_name:
+    #     search_path = os.getcwd()
+    # else:
+    #     search_path = os.path.normpath(f"{cwd}/{source_folder_name}/**")
+    # all_paths = glob.glob(search_path, recursive=True)
+    # print(all_paths)
+    # logger.debug(f"Found {len(all_paths)}.")
+    # logger.debug(all_paths)
+    # return remove_directories_from_path_list(all_paths)
+    if source_folder_name is None:
```
johnathan-rodriguez, line 245: needs to handle `""` and `None` I would simplify this to something like source_folder_name = source_folder_name or os.getcwd()
or if not source_folder_name:
     source_folder_name = os.getcwd()

shipyard_blueprints/shipyard-bp-utils/shipyard_bp_utils/files.py
johnathan-rodriguez: personally the following is more readable but I don't have a strong opinion other than using a var name for the full path assuming the full path doesn't break the other functions using this method.

```
all_files(pick a better var name) = os.listdir(source_folder_name)
file_list=[]
for file in all_files:
    full_path  = os.path.join(source_folder_name,file)
    if os.path.isfile(full_path):
        file_list.append(full_path
```

johnathan-rodriguez (commented):
confirm that the change doesn't break the other functions using find_all_local_files

johnathan-rodriguez (approved):

PR #166, opened 2024-02-22
Updated lock file and removed the path dependencies as a precaution

johnathan-rodriguez:


wrp801 (changes requested):
This also needs a version bump and a re-publish I believe

wrp801 (approved):

PR #173, opened 2024-03-01
Refactored SFTP

johnathan-rodriguez:


shipyard_blueprints/sftp/shipyard_sftp/__init__.py
```diff
@@ -1 +1 @@
-from shipyard_sftp.cli.sftp import SftpClient
```
johnathan-rodriguez, line 1: Remove because not needed

shipyard_blueprints/sftp/shipyard_sftp/sftp.py
```diff
@@ -0,0 +1,364 @@
...
+            paramiko.AuthenticationException,
+            ValueError,
+        ) as auth_error:
+            raise ExitCodeException(
+                auth_error, self.EXIT_CODE_INVALID_CREDENTIALS
+            ) from auth_error
+
+        except Exception as err:
+            raise ExitCodeException(
+                f"Error accessing the SFTP server. Server response:{err.__class__.__name__}:{err} ",
```
wrp801, line 279: Same comment as above

shipyard_blueprints/sftp/shipyard_sftp/cli/move_file.py
```diff
@@ -41,177 +46,78 @@ def get_args():
...
+            args.destination_folder_name
         )
-        print(f"{len(matching_file_names)} files found. Preparing to move...")
+        source_file_name_match_type = args.source_file_name_match_type
 
-        for index, key_name in enumerate(matching_file_names, 1):
-            destination_full_path = shipyard.files.determine_destination_full_path(
+        if source_file_name_match_type in [
+            "exact_match",
+            "",
```
wrp801, line 85: I don't think the "" is necessary, given the condition:
```    parser.add_argument(
        "--source-file-name-match-type",
        dest="source_file_name_match_type",
        choices={"exact_match", "regex_match"},
        default="exact_match",
        required=False,
        ```

File: shipyard_blueprints/sftp/shipyard_sftp/cli/delete_file.py [hunk 4]
```diff
@@ -29,129 +33,75 @@ def get_args():
...
 
-    if source_file_name_match_type == "regex_match":
-        files = find_sftp_file_names(client=client, prefix=source_folder_name)
-        matching_file_names = shipyard.files.find_all_file_matches(
-            files, re.compile(source_file_name)
+    try:
+        args = get_args()
+        if not args.password and not args.key:
+            raise ExitCodeException(
+                "Must specify a password or a RSA key",
```
wrp801, line 43: Instead of RSA key put 'Private Key File'

shipyard_blueprints/sftp/shipyard_sftp/cli/delete_file.py
```diff
@@ -29,129 +33,75 @@ def get_args():
...
+
+                logger.info(f"Deleting file {index} of {len(matching_file_names)}")
+                try:
+                    sftp.remove(delete_file_path)
+                except Exception as e:
+                    print(f"Failed to delete {file_name} due to {e}... Skipping")
+        elif source_file_name_match_type == "exact_match":
+            sftp.remove(source_full_path)
+    except ExitCodeException as e:
+        logger.error(e)
```
wrp801, line 92: Should be e.message

shipyard_blueprints/sftp/shipyard_sftp/cli/delete_file.py
```diff
@@ -29,129 +33,75 @@ def get_args():
...
+        exit(e.exit_code)
+    except Exception as e:
+        logger.error(e)
```
wrp801, line 95: do `str(e)`

wrp801 (changes requested):

shipyard_blueprints/sftp/shipyard_sftp/cli/authtest.py
```diff
@@ -1,23 +1,23 @@
...
 
 
 def main():
-    sftp = SftpClient(
-        host=os.getenv("SFTP_HOST"),
-        port=os.getenv("SFTP_PORT"),
-        user=os.getenv("SFTP_USERNAME"),
-        pwd=os.getenv("SFTP_PASSWORD"),
-        key=None,
+    sys.exit(
```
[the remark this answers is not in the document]
johnathan-rodriguez: I was under the impression that private keys are not currently supported for the credential tests so adding them to the credential group wouldn't be a viable option as the application exists today

shipyard_blueprints/sftp/shipyard_sftp/sftp.py
```diff
@@ -0,0 +1,364 @@
...
+import re
+import stat
+
+import paramiko
+from shipyard_templates import CloudStorage, ExitCodeException
+from shipyard_templates.shipyard_logger import ShipyardLogger
+
+logger = ShipyardLogger().get_logger()
+
+DEFAULT_WINDOW_SIZE = 4294967294
```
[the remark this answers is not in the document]
johnathan-rodriguez: There was an issue with sessions timing out. This helped throttle the larger upload and downloads

shipyard_blueprints/sftp/shipyard_sftp/sftp.py
```diff
@@ -0,0 +1,364 @@
...
+            raise ExitCodeException(
+                f"Error: the file {filename} was not not found",
+                self.EXIT_CODE_FILE_NOT_FOUND,
+            ) from e
+        except ExitCodeException as error:
+            raise error
+        except Exception as e:
+            raise ExitCodeException(
+                f"Failed to download {filename} to {destination}. "
+                f"Response from server: {e.__class__.__name__}: {e}",
```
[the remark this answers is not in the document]
johnathan-rodriguez: `{str(e)}` is not needed. `{e}` prints out the string representation of the exception. I included `{e.__class__.__name__}:` to help handle future unknown errors. The output would look something like

`Failed to download test.csv to source/test.csv. Response from server: ValueError : port cannot be a string must be an integer`

So in the future if that is an error that happens often we can then do something like
```
try:
     ...
 except ValueError as e:
      raise ExitCodeException("The invalid port was used. Update credentials". self.EXIT_CODE..
 except ExitCodeException
     ...
 except Exception:
     ....
```

you can test locally that the handled exceptions do not throw a full trace back by running 
```
    def raise_test_exception():
        raise ExitCodeException("Test exception", 1)

    try:
        raise_test_exception()
        print(e)
```

can also modify the above to look at the output of e.__class__.__name__

shipyard_blueprints/sftp/shipyard_sftp/cli/move_file.py
johnathan-rodriguez: That is to handle how the application sends empty values. Essentially enforcing the default value

shipyard_blueprints/sftp/shipyard_sftp/cli/delete_file.py
johnathan-rodriguez: Similar to above comment. (e) and (e.message) essentially log the same output

shipyard_blueprints/sftp/shipyard_sftp/cli/delete_file.py
johnathan-rodriguez: same response as before.

shipyard_blueprints/sftp/shipyard_sftp/cli/authtest.py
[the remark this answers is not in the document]
wrp801: Correct it is not supported. What I'm saying is that if we wish to have a credential test for these parameters (and not the private key file), we need to make it clear that even though these may be valid parameters, a private key file is likely necessary to connect to an SFTP server

shipyard_blueprints/sftp/shipyard_sftp/cli/move_file.py
wrp801: It's fine to leave it in, but in the application this would be a `Select` value and those always have a default option

wrp801 (approved):

PR #193, opened 2024-03-19
Wes/integ 311 rewriterefactor

wrp801:


shipyard_blueprints/postgresql/shipyard_postgresql/cli/execute_sql.py
```diff
@@ -14,21 +19,9 @@ def get_args():
     parser.add_argument("--database", dest="database", required=False)
     parser.add_argument("--port", dest="port", default="5432", required=False)
     parser.add_argument("--url-parameters", dest="url_parameters", required=False)
-    parser.add_argument("--db-connection-url", dest="db_connection_url", required=False)
```
johnathan-rodriguez, line 17: I am assuming this is not an input in the current blueprint,correct?

shipyard_blueprints/postgresql/shipyard_postgresql/cli/execute_sql.py
```diff
@@ -39,53 +32,34 @@ def get_args():
...
     except Exception as e:
-        print(f"Failed to execute query {args.query} due to {e}")
+        logger.error(
+            f"An unexpected error occurred when attempting to execute query in Postgres. Message from the server reads: {e}"
+        )
+        sys.exit(Database.EXIT_CODE_UNKNOWN)
+
     finally:
-        session.close()
+        postgres.close()
```
johnathan-rodriguez, line 62: `postgres` may have to be initialized so that you can do a `if postrgres: postrgres.close` otherise this could throw a confusing traceback error for the customer

shipyard_blueprints/postgresql/shipyard_postgresql/cli/store_query_results.py
```diff
@@ -52,103 +44,46 @@ def get_args():
...
+        target_path = shipyard.files.combine_folder_and_file_name(
+            folder_name=target_dir, file_name=target_file
         )
+        file_header = shipyard.args.convert_to_boolean(args.file_header)
+        query = text(args.query)
 
-    db_string = os.environ.get("DB_CONNECTION_URL")
-    return db_string
-
+        # args for client
```
johnathan-rodriguez, line 60: remove comment. the var name is descriptive enough

shipyard_blueprints/postgresql/shipyard_postgresql/cli/store_query_results.py
```diff
@@ -52,103 +44,46 @@ def get_args():
...
 
-    create_csv(
-        query=query,
-        db_connection=db_connection,
-        destination_file_path=destination_full_path,
-        file_header=file_header,
-    )
-    db_connection.dispose()
+    finally:
+        postgres.close()
```
johnathan-rodriguez, line 86: same comment re: initializing the postrgres client as a precaution

shipyard_blueprints/postgresql/shipyard_postgresql/cli/upload_file.py
```diff
@@ -65,182 +58,75 @@ def get_args():
...
+        match_type = args.source_file_name_match_type
+        src_file = args.source_file_name
+        src_dir = args.source_folder_name
+        src_path = shipyard.files.combine_folder_and_file_name(
+            folder_name=src_dir, file_name=src_file
         )
+        table_name = args.table_name
+        insert_method = args.insert_method
+
+        # args for client
```
johnathan-rodriguez, line 73: remove comment

shipyard_blueprints/postgresql/shipyard_postgresql/cli/upload_file.py
```diff
@@ -65,182 +58,75 @@ def get_args():
...
-
-            chunk.to_sql(
-                table_name,
-                con=db_connection,
-                index=False,
-                if_exists=insert_method,
-                method=upload_method,
-                chunksize=10000,
-                schema=schema,
+        else:
```
johnathan-rodriguez, line 106: I think its best to be explicit when possible
```
if match_type == exact_match:
elif match_type == regex
```

something i care less about but I have been doing in the scripts I refactored is moving the exact match logic before the regex since it doesn't have the nested logic that regex has.

shipyard_blueprints/postgresql/shipyard_postgresql/cli/upload_file.py
```diff
@@ -65,182 +58,75 @@ def get_args():
...
-        upload_data(
-            source_full_path=source_full_path,
-            table_name=table_name,
-            insert_method=insert_method,
-            db_connection=db_connection,
-            schema=schema,
-        )
-    db_connection.dispose()
+    finally:
+        postgres.close()
```
johnathan-rodriguez, line 129: Same comment here about initializing the progress var

shipyard_blueprints/postgresql/shipyard_postgresql/errors/exceptions.py
```diff
@@ -0,0 +1,23 @@
+from logging import error
```
johnathan-rodriguez, line 1: Not seeing this being used

shipyard_blueprints/postgresql/shipyard_postgresql/errors/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+from shipyard_templates import Database, ExitCodeException
+
```
johnathan-rodriguez, line 3: should move outside of the errors folder

shipyard_blueprints/postgresql/shipyard_postgresql/errors/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+EXIT_CODE_UPLOAD_ERROR = 101
+EXIT_CODE_CHUNK_DOWNLOAD_ERROR = 102
+
+
+class InvalidCredentials(ExitCodeException):
```
johnathan-rodriguez, line 8: If you want to follow the benefit of adding common exceptions to the abstract class this should be move to the database class

shipyard_blueprints/postgresql/shipyard_postgresql/errors/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, error_message: Exception):
+        self.message = f"Error in connecting to Postgres. Message from the server reads: {error_message}"
+        self.exit_code = Database.EXIT_CODE_INVALID_CREDENTIALS
+
+
+class PostgresUploadError(ExitCodeException):
```
johnathan-rodriguez, line 14: Should use the database class otherwise the template abstract class exceptions don't really add value.

shipyard_blueprints/postgresql/shipyard_postgresql/errors/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, error_message: Exception):
+        self.message = f"Error attempting to load data into Postgres. Message from the server reads: {error_message}"
+        self.exit_code = EXIT_CODE_UPLOAD_ERROR
+
+
+class ChunkDownloadError(ExitCodeException):
```
johnathan-rodriguez, line 20: Same comment applies here

shipyard_blueprints/postgresql/shipyard_postgresql/postgres.py
```diff
@@ -1,15 +1,26 @@
-from shipyard_templates import Database
```
johnathan-rodriguez, line 1: The overall comment to the below code is the error handling is very broad. Ideally a there should be handling of more specific errors

johnathan-rodriguez (changes requested):

shipyard_blueprints/postgresql/shipyard_postgresql/cli/execute_sql.py
wrp801: correct

shipyard_blueprints/postgresql/shipyard_postgresql/postgres.py
wrp801: Interested to know your thoughts here and possible improvements. There are (in my opinion) too many possible errors to handle individually (https://docs.sqlalchemy.org/en/20/errors.html#connections-and-transactions) so I opted for more generalized error handling relying on the error message (which in my experience is often informative enough). Each method more or less has its own custom exception which spits out the message from API. I've tried to strike the balance between pragmatism and utility for end user, so if I've missed something here or you have possible improvements, I'd very much welcome that because it would not only apply to this but also to the other database blueprints

shipyard_blueprints/postgresql/shipyard_postgresql/cli/execute_sql.py
```diff
@@ -39,53 +28,34 @@ def get_args():
...
-    Session = sessionmaker(bind=db_connection)
-    return db_connection, Session
-
-
 def main():
     args = get_args()
     query = text(args.query)
-
-    db_string = create_connection_string(args)
+    client_args = {
```
johnathan-rodriguez, line 34: Sorry if I wasn't clear. postgres = None should be the only thing outside of the try block.

end goal is to avoid any risk of a `NameError` in the finally clause. 

```
    client_args = {
        "user": args.username,
        "pwd": args.password,
        "host": args.host,
        "database": args.database,
        "port": args.port,
        "url_params": args.url_parameters if args.url_parameters != "" else None,
    }
    postgres = PostgresClient(**client_args)
    ```
    has a chance of failing and if it fails outside of the try block it will throw a traceback error

johnathan-rodriguez (changes requested):
Small change re: the client var

johnathan-rodriguez (approved):

PR #225, opened 2024-04-02
Wes/integ 405 portable trigger flow

wrp801:


shipyard_blueprints/portable/shipyard_portable/errors.py
```diff
@@ -0,0 +1,15 @@
+from shipyard_templates import ExitCodeException, Etl
+
```
johnathan-rodriguez, line 2: rename folder to exceptions to match the newer format and move common exceptions to abstract class

johnathan-rodriguez: in this case looks like both exceptions can be moved to the abstract class

shipyard_blueprints/portable/shipyard_portable/portable.py
```diff
@@ -0,0 +1,152 @@
...
+                logger.debug("Flow status is: pending")
+                return Etl.EXIT_CODE_FINAL_STATUS_PENDING
+            elif status == "NOT_RUNNING":
+                logger.debug("Flow status is: not running")
+                return Etl.EXIT_CODE_FINAL_STATUS_INCOMPLETE
+            else:
+                logger.error(f"Unknown status: {status}")
+                return Etl.EXIT_CODE_UNKNOWN_ERROR
+        else:
+            logger.error(f"Unknown disposition: {disposition}")
```
johnathan-rodriguez, line 151: Should this raise an exception?

shipyard_blueprints/portable/shipyard_portable/portable.py
```diff
@@ -0,0 +1,152 @@
...
+                logger.debug("Flow status is: still running")
+                return Etl.EXIT_CODE_SYNC_ALREADY_RUNNING
+            elif status == "PENDING":
+                logger.debug("Flow status is: pending")
+                return Etl.EXIT_CODE_FINAL_STATUS_PENDING
+            elif status == "NOT_RUNNING":
+                logger.debug("Flow status is: not running")
+                return Etl.EXIT_CODE_FINAL_STATUS_INCOMPLETE
+            else:
+                logger.error(f"Unknown status: {status}")
```
johnathan-rodriguez, line 148: should this raise an exception

shipyard_blueprints/portable/shipyard_portable/portable.py
```diff
@@ -0,0 +1,152 @@
...
+        disposition = response_data.get("disposition")
+        status = response_data.get("lifecycle")
+
+        if disposition == "SUCCEEDED":
+            logger.debug("Flow completed successfully")
+            return Etl.EXIT_CODE_FINAL_STATUS_COMPLETED
+
+        elif disposition == "FAILED":
+            logger.error("Flow failed")
+            return Etl.EXIT_CODE_FINAL_STATUS_ERRORED
```
johnathan-rodriguez, line 136: for consistency `self.EXIT_CODE_FINAL_STATUS_ERRORED`

shipyard_blueprints/portable/tests/test_auth.py
```diff
@@ -0,0 +1,37 @@
+import os
+import subprocess
+import pytest
+from shipyard_portable import PortableClient
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
```
johnathan-rodriguez, line 7: should add a clause to skip tests if env file is not found

shipyard_blueprints/portable/tests/test_auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_authtest_good_connection(auth):
+    result = subprocess.run(auth, capture_output=True)
+    assert result.returncode == 0
+
+
+def test_authtest_bad_connection(auth):
+    original = os.getenv("PORTABLE_API_TOKEN")
+    os.environ["PORTABLE_API_TOKEN"] = "bad_token"  # set bad token temporarily
```
johnathan-rodriguez, line 34: no change needed to be made but you can use `monkeypatch.setenv(` for this function.

shipyard_blueprints/portable/shipyard_portable/portable.py
```diff
@@ -0,0 +1,152 @@
...
+            Dict[str, str]: A dictionary containing the synchronization status.
+
+        Raises:
+            UnauthorizedError: If the request is unauthorized (status code 401).
+            BadRequestError: If the request is invalid (status code 400).
+            ExitCodeException: If an unexpected error occurs while fetching the status.
+        """
+        try:
+            flow_url = f"{self.flow_url}/{flow_id}/status"
+            response = requests.get(flow_url, headers=self.headers)
```
johnathan-rodriguez, line 96: suggestions:
flow_url is not reused so I would've
`response = requests.get(f"{self.flow_url}/{flow_id}/status", headers=self.headers)`

also would've made a `_handle_error` function since 
```
  except Exception as e:
            if response.status_code == 401:
                raise UnauthorizedError(response.text)
            elif response.status_code == 400:
                raise BadRequestError(response.text)
            else:
```
seems to be reused. It is a small code so not needed at this point but still worth considering

johnathan-rodriguez (changes requested):

shipyard_blueprints/portable/tests/test_auth.py
wrp801: I like this way more than manually setting the variables

shipyard-templates/shipyard_templates/etl.py
```diff
@@ -1,4 +1,17 @@
 from abc import ABC, abstractmethod
+from .exit_code_exception import ExitCodeException
+
+
+class UnauthorizedError(ExitCodeException):
+    def __init__(self, message: str):
+        self.message = f"Failed to connect to the API. Ensure that the API key provided is correct. Message from the API: {message}"
```
johnathan-rodriguez, line 7: small tweak here:
instead of `API key` you should say credentials to cast a wider net

shipyard_blueprints/portable/shipyard_portable/cli/trigger.py
```diff
@@ -0,0 +1,65 @@
...
+        portable = PortableClient(args.access_token)
+        portable.connect()
+        flow_response = portable.trigger_sync(args.flow_id)
+
+        wait = shipyard.args.convert_to_boolean(args.wait_for_completion)
+        flow_status = portable.get_sync_status(flow_id)
+        flow_disposition = flow_status["disposition"]
+        exit_code = portable.determine_sync_status(flow_status)
+        if wait:
+            logger.info("The flow will be checked every 60 seconds until completion")
```
johnathan-rodriguez, line 40: logger.info("The flow will be checked every 60 seconds until completion")
should be             logger.info(f"The flow will be checked every{INTERVAL} seconds until completion")

shipyard_blueprints/portable/shipyard_portable/cli/trigger.py
```diff
@@ -0,0 +1,65 @@
...
+        logger.info(f"Flow has a status of {flow_disposition}, exiting now")
+
+    except ExitCodeException as ec:
+        logger.error(ec.message)
+        sys.exit(ec.exit_code)
+    except Exception as e:
+        logger.error(f"An unexpected error occurred: {e}")
+        sys.exit(Etl.EXIT_CODE_UNKNOWN_ERROR)
+    else:
+        Artifact("portable").responses.write_json("flow_response.json", flow_status)
```
johnathan-rodriguez, line 60: write_json may append `.json` so this may save as "flow_response.json.json". I will double check this later this morning and get back to you

johnathan-rodriguez (changes requested):
Small changes for logging

shipyard_blueprints/portable/shipyard_portable/cli/trigger.py
johnathan-rodriguez: image

it does append .json so file name should be without the extension

johnathan-rodriguez (changes requested):
fix artifact filename

wrp801:
Going to bump the version of templates and portable out of pre-release

PR #229, opened 2024-04-04
Can set transport for client connection

johnathan-rodriguez:


shipyard_blueprints/sftp/shipyard_sftp/sftp.py
```diff
@@ -18,16 +18,14 @@
...
-DEFAULT_REKEY_PACKETS = pow(2, 40)
+
 
 
 class SftpClient(CloudStorage):
     EXIT_CODE_DELETE_ERROR = 100
 
     def __init__(
-            self, host: str, port: int, key: str = None, user: str = None, pwd: str = None
+            self, host: str, port: int, key: str = None, user: str = None, pwd: str = None, transport=None
```
wrp801, line 28: Not to be overly pedantic, but two points here:
- Whenever something is initialized to None it should be `Optional[type]`
- For the library code in particular, we should have all types annotated. The `transport` variable should be of type `paramiko.Transport`

wrp801 (changes requested):
You're going to think I'm a stickler after this

wrp801 (approved):

PR #242, opened 2024-04-04
Jr/integ 286 rewriterefactor

johnathan-rodriguez:


johnathan-rodriguez:
my ide is doing that fun thing again when it is including more files than it shows in  git status. I will make a new branch in a moment- let me know if you think it is worth handling the exceptions on this blueprint. the original does direct exits with exit code ranging 1-10

wrp801 (changes requested):
It looks like you may have gotten mixed up in your directories and bumped the version of googledrive instead. Remove the GoogleDrive and Smartsheets data then repush

PR #245, opened 2024-04-05
Bug fix/sftp backwards compatible

johnathan-rodriguez:


johnathan-rodriguez:
This is a stop gap for now to make sure no fleets continue to have issues over the weekend. We should discuss how would like to handle when key and password inputs are provided. ie should it raise an error, should it default to a method(currently how it is being handled), or fallback so it attempts both credential types

shipyard_blueprints/sftp/shipyard_sftp/cli/download_file.py
```diff
@@ -64,9 +64,8 @@ def main():
             transport.default_window_size = DEFAULT_WINDOW_SIZE
             transport.packetizer.REKEY_BYTES = DEFAULT_REKEY_BYTES
             transport.packetizer.REKEY_PACKETS = DEFAULT_REKEY_PACKETS
-            transport.connect(None, args.username, args.password)
-
             connection_args["transport"] = transport
```
wrp801, line 67: does this need to be connection_args.get("transport") instead to ensure that it defaults to `None`? Or does it have a default value in the CLI?

wrp801 (approved):

PR #299, opened 2024-04-18
added script portion to authtest

wrp801:


shipyard_blueprints/motherduck/shipyard_motherduck/cli/authtest.py
```diff
@@ -11,10 +11,15 @@ def main():
         print(f"Token is {os.getenv('MOTHERDUCK_TOKEN')}")
```
johnathan-rodriguez, line 11: I missed this in the previous review but this should be removed

johnathan-rodriguez (changes requested):
rem

johnathan-rodriguez (approved):

PR #317, opened 2024-05-02
Wes/integ 448 handle user agent in snowflake credential test

wrp801:
Left some notes for future refactoring needed for snowflake class, specifically the connect method and the exceptions


johnathan-rodriguez (commented):
The authentication test works fine right now, and I noticed you plan to improve it later  but might as well do it now while you're already in there.

If you still feel it's better to delay these changes, I'm will approve the pr to keep things moving.

johnathan-rodriguez (approved):

PR #324, opened 2024-05-14
Updated logger and shipyard-templates version for Fivetran

johnathan-rodriguez:


shipyard_blueprints/fivetran/pyproject.toml
```diff
@@ -11,13 +11,13 @@ packages = [{include = "shipyard_fivetran"}]
 python = "^3.9"
 pytz = "2023.3"
 requests = "2.31.0"
-shipyard-templates = "0.1.6"
 shipyard-utils = "0.1.4"
 
 
 #[tool.poetry.group.dev.dependencies]
 #shipyard-bp-utils = {path = "../shipyard-bp-utils",develop=true}
```
wrp801, line 18: You can remove the commented out section

shipyard_blueprints/fivetran/shipyard_fivetran/cli/update_connector.py
```diff
@@ -61,7 +63,7 @@ def main():
             **args_dict,
         )
     except ExitCodeException as e:
-        fivetran_client.logger.error(e)
```
wrp801, line 64: Instead of just logging the exception, put a message in there like we do for the others (i.e. "Something went wrong when attempting to do xyz, message from the server/api: {e}")

wrp801 (changes requested):
Small changes

shipyard_blueprints/fivetran/shipyard_fivetran/cli/update_connector.py
johnathan-rodriguez: Tried making minimal code change in order to introduce breaking changes.

There is already a line that logs
`logger.error(f"Error: {resp.status_code} - {resp.text}")` 

and the exception that should get logged when an unknown error occurs is 
`raise ExitCodeException(f"{resp.json()}", self.EXIT_CODE_UNKNOWN_ERROR)`

The others use `.get` which in theory could return None so I think it is best to not use the typical "Message from server"

but I did wrap the error message with some english to hopefully make it easier for the end user to follow

PR #325, opened 2024-05-14
Added upsert and enable trigger_fleet to be ran with overrides

johnathan-rodriguez:


shipyard_blueprints/shipyard/shipyard_api/client.py
```diff
@@ -139,33 +188,111 @@ def export_voyages(self, target_path: str):
...
+                                 headers={
+                                     "accept": "application/json",
+                                     "content-type": "application/json",
+                                     "X-Shipyard-API-Key": self.api_key
+                                 }
+                                 )
+
+        log_data = response.content.decode('utf-8')
+
+        df = pandas.read_csv(io.StringIO(log_data))
```
wrp801, line 296: You should include an try/except here to handle cases where the `run_id` does not exist in dataframe. If you leave it as is, this will throw an `IndexError: index 0 is out of bounds for axis 0 with size 0`

wrp801 (changes requested):

shipyard_blueprints/shipyard/shipyard_api/client.py
```diff
@@ -139,33 +189,120 @@ def export_voyages(self, target_path: str):
...
+
+            log_data = response.content.decode("utf-8")
+
+            df = pandas.read_csv(io.StringIO(log_data))
+            fleet_run = df[df["Fleet Log ID"] == run_id]
+            return fleet_run["Status"].values[0]
+        except ExitCodeException:
+            raise
+        except Exception as e:
+            raise ExitCodeException(
```
wrp801, line 305: Technically this is fine. You could be more specific and catch the IndexOutOfBounds exception and say `The provided run ID is not found in any of the runs` or something along those lines. Not necessary, but just food for thought

wrp801 (approved):

PR #326, opened 2024-05-15
Jr/integ 456 sftp rsa key credential type test

johnathan-rodriguez:


shipyard_blueprints/sftp/shipyard_sftp/sftp.py
```diff
@@ -260,9 +261,16 @@ def get_sftp_client(self) -> paramiko.SFTPClient:
                 logger.debug(
                     f"Connecting to {self.host} using key-based authentication"
                 )
+
                 ssh = paramiko.SSHClient()
                 ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-                key = paramiko.RSAKey.from_private_key_file(self.key)
+                if self.key.startswith("-----BEGIN"):
```
wrp801, line 267: Is there no other way to determine if it is a private key file?

johnathan-rodriguez: I had also considered:
1. os.path.isfile()
2. os.path.isdir()
3. os.path.exists()
4. regex match

johnathan-rodriguez: another option is to try 1 method and then fallback to the other. When you try to use `from_private_key_file` with the string value we get an error . str does not have method readlines

PR #330, opened 2024-05-20
Jr/integ 459 credential testing

johnathan-rodriguez:


shipyard_blueprints/dbt/shipyard_dbt/cli/authtest.py
```diff
@@ -8,7 +8,7 @@ def main():
     sys.exit(
         DbtClient(
             access_token=os.getenv("DBT_API_KEY"),
-            account_id=os.getenv("DBT_ACCOUNT_ID"),
+            account_id='',
```
wrp801, line 11: No change here, as we discussed

shipyard_blueprints/s3/shipyard_s3/cli/authtest.py
```diff
@@ -12,8 +12,7 @@ def main():
     sys.exit(
         S3Client(
             aws_access_key=os.getenv("AWS_ACCESS_KEY_ID"),
-            aws_secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
-            region=os.getenv("AWS_DEFAULT_REGION"),
```
wrp801, line 16: We need to leave the region as a required field here

wrp801 (changes requested):
Main change is for S3, we need to have the region as a required credential field.

shipyard_blueprints/s3/shipyard_s3/cli/authtest.py
johnathan-rodriguez: currently not being used in prod
image

wrp801 (approved):

PR #338, opened 2024-05-23
Update package to use ShipyardLogger

johnathan-rodriguez:


shipyard_blueprints/trello/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "shipyard-trello"
-version = "0.1.1"
+version = "0.1.2a0"
```
johnathan-rodriguez, line 3: leaving as a pre release for now since Trello seems to be having some intermittent issues on their servers

PR #339, opened 2024-05-23
Fix client init bug and authtest

johnathan-rodriguez:


johnathan-rodriguez:
I believe databricks may have been missed during the refactor project but for now this is only addressing what is needed for credentials

PR #356, opened 2024-07-15
Jr/integ 478 create blueprint class

johnathan-rodriguez:


shipyard_blueprints/azure-data-factory/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
+name = "shipyard-azure-data-factory"
+version = "0.1.0a0"
+description = ""
```
wrp801, line 4: Add a short description for this

shipyard_blueprints/azure-data-factory/shipyard_azure_data_factory/azure_data_factory.py
```diff
@@ -0,0 +1,96 @@
...
+
+from azure.identity import ClientSecretCredential
+from azure.mgmt.datafactory import DataFactoryManagementClient
+from shipyard_templates import Etl, ShipyardLogger, ExitCodeException
+
+logger = ShipyardLogger.get_logger()
+
+
+class AzureDataFactoryClient(Etl):
+    def __init__(self, client_id, client_secret, tenant_id, subscription_id):
```
wrp801, line 11: Add type annotations for these variables

shipyard_blueprints/azure-data-factory/shipyard_azure_data_factory/azure_data_factory.py
```diff
@@ -0,0 +1,96 @@
...
+        :param tenant_id: Azure tenant ID
+        :param subscription_id: Azure subscription ID
+        """
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.tenant_id = tenant_id
+        self.subscription_id = subscription_id
+        self.adf_client = None
+
+    def setup_client(self):
```
wrp801, line 26: Since you are referencing the `self.adf_client` in the `trigger_sync` method, and doing a check that it exists, I would recommend setting a property for it that calls this setup_client. Something like: 

```python
def __init__(...):
    ... 
    self._adf_client = None

@property 
def adf_client(self):
    if not self.adf_client:
        self._adf_client = self.setup()
    return self._adf_client
```
That was the approach I used in the databases and it worked pretty well

shipyard_blueprints/azure-data-factory/shipyard_azure_data_factory/azure_data_factory.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        if not self.adf_client:
+            self.setup_client()
+
+            run_response = self.adf_client.pipelines.create_run(resource_group, data_factory_name, pipeline_name,
+                                                                parameters={})
+            logger.info(f"Triggered pipeline run with ID: {run_response.run_id}")
+            if wait_for_completion:
+                run_status = self.determine_sync_status(resource_group, data_factory_name, run_response.run_id)
+                while run_status not in {"Succeeded", "Failed", "Cancelled", "Canceling"}:
```
wrp801, line 78: I'm assuming these are the official statues from ADF?

wrp801 (changes requested):
Small changes to make

shipyard_blueprints/azure-data-factory/shipyard_azure_data_factory/azure_data_factory.py
johnathan-rodriguez: the statuses I see are 
Queued, InProgress, Succeeded, Failed, Canceling, Cancelled
listed here:
https://learn.microsoft.com/en-us/rest/api/datafactory/pipeline-runs/get?view=rest-datafactory-2018-06-01&tabs=HTTP#pipelinerun

shipyard_blueprints/azure-data-factory/shipyard_azure_data_factory/azure_data_factory.py
johnathan-rodriguez: Agreed- I think this should be the standard. So push me again if I ever forget again

wrp801 (approved):

PR #360, opened 2024-08-01
Adjusted for OAuth user flow support

johnathan-rodriguez:


johnathan-rodriguez:
Simple username and password is also an option that I have added support for- This is discouraged by Microsoft because it is not as secure as the other methods but it is an option. Commented out the code that would've handled setting the credential group because args.username args.password are not valid options with how the blueprints work now.

shipyard_blueprints/microsoft-onedrive/shipyard_microsoft_onedrive/utils.py
```diff
@@ -0,0 +1,29 @@
...
+        logger.debug("Using access token for authentication")
+        return {"access_token": args.access_token}
+    if args.client_id and args.client_secret and args.tenant:
+        logger.debug("Using Client ID, Client Secret, and Tenant for authentication")
+        return {
+            "client_id": args.client_id,
+            "client_secret": args.client_secret,
+            "tenant": args.tenant,
+        }
+    # if args.client_id and args.tenant and args.username and args.password:
```
johnathan-rodriguez, line 17: Currently not supported in our blueprints but added here for an easy patch when we roll out the credential group support

shipyard_blueprints/microsoft-onedrive/shipyard_microsoft_onedrive/drive_client.py
```diff
@@ -187,43 +247,33 @@ def move(
...
             drive_id: The ID of the drive to remove (only necessary if using basic authentication)
         """
-        url = f"https://graph.microsoft.com/v1.0/drives/{drive_id}/root:/{drive_path}"
-        headers = {
-            "Authorization": f"Bearer {self.access_token}",
-        }
-
-        response = requests.delete(url, headers=headers)
+        url = f"{BASE_URL}/drives/{drive_id}/root:/{drive_path}"
+        response = request(
```
wrp801, line 270: shouldn't this be `self._request` instead?

shipyard_blueprints/microsoft-onedrive/shipyard_microsoft_onedrive/drive_client.py
```diff
@@ -246,139 +296,121 @@ def get_folder_id(self, folder_name: str, drive_id: Optional[str]) -> Optional[s
...
-            file_name: The name of the file to write to
+            download_url (str): The download URL from the Graph API.
+            file_name (str): The name of the file to write to.
 
         Raises:
-            BadRequestError:
+            ExitCodeException: If the download fails.
         """
-        response = requests.get(download_url)
+        response = request("GET", download_url)
```
wrp801, line 413: I would have just left it as `requests.get`

wrp801 (changes requested):
Mostly small changes and addition of docstrings

shipyard_blueprints/microsoft-onedrive/shipyard_microsoft_onedrive/drive_client.py
johnathan-rodriguez: I changed how the module was imported so this was changed to reflect that

wrp801 (approved):

PR #362, opened 2024-08-07
Microsoft Sharepoint: OAuth Support

johnathan-rodriguez:
* refactored the class to use an access token property that can handle multiple credential formats. 
* Refactored the upload cli to use the shipyard-bp-utils functions for consistent behavior between blueprints
* Other than the above the no code change was done to the clis other than using a utils for retrieving credentials

**Note: I will bump the version once we are able to fully test this in the application- as of now tests are successful locally


shipyard_blueprints/microsoft-sharepoint/shipyard_microsoft_sharepoint/utils.py
```diff
@@ -0,0 +1,29 @@
+from shipyard_templates import ShipyardLogger, ExitCodeException, CloudStorage
+
+logger = ShipyardLogger.get_logger()
+
+
+def get_credential_group(args):
+    if args.access_token:
```
johnathan-rodriguez, line 7: this is most likely going to have to be adjusted to be `if access_token:=os.getenv("OAUTH_ACCESS_TOKEN"): return {"access_token":access_token}` but since the blueprint already had a flag for the access_token figured it would be worth attempting.

wrp801 (approved):
I think the only thing needed is to bump the version in the toml

PR #363, opened 2024-08-07
Wes/integ 540 bigquery oauth update

wrp801:


shipyard_blueprints/bigquery/shipyard_bigquery/bigquery.py
```diff
@@ -20,26 +20,52 @@
...
+                self.access_token,
+                scopes=[
+                    "https://www.googleapis.com/auth/bigquery",
+                    "https://www.googleapis.com/auth/bigquery.insertdata",
+                    "https://www.googleapis.com/auth/devstorage.read_write",
+                ],
+            )
         return service_account.Credentials.from_service_account_info(self.json_creds)
 
     @property
```
johnathan-rodriguez, line 51: is this property used/needed?

if not I would wrap this in a try clause then log a warning if the email is unable to be retrieved so that this doesn't throw a confusing error for the end user

johnathan-rodriguez (approved):
A very minor consideration re: email property but approved regardless of which path you end up taking

shipyard_blueprints/bigquery/shipyard_bigquery/bigquery.py
wrp801: It is needed, particularly in the credential test and some of the blueprints

PR #370, opened 2024-12-06
Handle bulk updates to campaign budgets and simplified arguments

johnathan-rodriguez:


shipyard_blueprints/magnite/shipyard_magnite/cli/update_campaign_budget.py
```diff
@@ -0,0 +1,133 @@
...
+        spend_profile = TargetingSpendProfile(id=profile_id, budgets=valid_budgets)
+        client.update_campaign_budgets(campaign_id, spend_profile)
+
+        if len(valid_budgets.items) == len(budgets.items):
+            logger.info(f"Successfully updated budgets for campaign ID: {campaign_id}")
+        else:
+            logger.warning(
+                f"Partial success: Updated budgets for campaign ID: {campaign_id}. \n"
+                "Note: Some budget items were invalid and excluded from the update."
+            )
```
forstmeier, line 64: Is this necessary here given the `<` check above?

shipyard_blueprints/magnite/shipyard_magnite/dataclasses/budget_item.py
```diff
@@ -0,0 +1,85 @@
...
+    "budget_period",
+    "budget_pacing",
+    "budget_metric",
+    "vast_caching_adjustment",
+    "budget_item_id",
+)
+
+
+@dataclass
+class BudgetItem:
```
forstmeier, line 30: Should this class have an `__init__` method with the `validate` method logic inside it?

shipyard_blueprints/magnite/shipyard_magnite/magnite.py
```diff
@@ -1,223 +1,163 @@
...
-    def _form_url(self, endpoint: str):
-        self.endpoint_url = f"{self.base_url}/{endpoint}"
+        campaign_data = self.get_campaign_by_id(id)
+        if file_type == "csv":
+            if isinstance(campaign_data, list):
+                df = pd.json_normalize(campaign_data)
+            else:
+                df = pd.json_normalize(
+                    [campaign_data]
+                )  # Wrap single dictionary in a list
```
forstmeier, line 133: Probably an unnecessary comment.

shipyard_blueprints/magnite/shipyard_magnite/utils.py
```diff
@@ -0,0 +1,335 @@
...
+    "net_cost",
+    "gross_cost",
+    "requests",
+    "revenue",
+    "impressions",
+    "completes",
+    "clicks",
+}
+VALID_BUDGET_PERIODS = {"day", "lifetime", "month", "week", "hour"}
+VALID_BUDGET_PACINGS = {"asap", "smooth", "front_loaded", "even"}
```
forstmeier, line 90: Are these identical to the other "constants" declared in the PR? If so, could they be consolidated somewhere?

shipyard_blueprints/magnite/shipyard_magnite/utils.py
```diff
@@ -0,0 +1,335 @@
...
+
+        if not _is_valid_budget_period(budget_item.get("budget_period")):
+            result["status"] = "error: invalid budget period"
+            logger.error(
+                f"Invalid budget period in item {budget_item.get('id')}. Skipping..."
+            )
+            item_results.append(result)
+            continue
+
+        if not _is_valid_budget_pacing(budget_item.get("budget_pacing")):
```
forstmeier, line 206: Should these be `self` methods invocations?

forstmeier (changes requested):

shipyard_blueprints/magnite/shipyard_magnite/utils.py
johnathan-rodriguez: I am going to opt to just remove the old code completely. Mainly kept it in in case there was a use case where using the dataclasses was too cumbersome

shipyard_blueprints/magnite/shipyard_magnite/dataclasses/budget_item.py
johnathan-rodriguez: If it weren’t for the requirement to allow partial successful uploads I would’ve agreed with you. I could have moved the validations into a `__post_init__` but since but since we're not forcing the data to be correct the ability to have the dataclass in an invalid state and mutable seemed more inline with the ask. Also made grouping the reporting easier by having the validation step be explicit. 

for context I expect there to be around 5 dataclasses as it relates to the Campaign so simpler is best in my opinion.

shipyard_blueprints/magnite/shipyard_magnite/cli/update_campaign_budget.py
johnathan-rodriguez: yes and no.

these logs happen after `client.update_campaign_budgets(campaign_id, spend_profile)` step so in theory something wrong can happen there. 

If I just had it as `Success` felt like it could be misleading to the end user that the update went off without issue. I wanted to make it very clear that yes the edit to the campaign did go through but not all of the budgets that were inputted made its way.

forstmeier (approved):

